PR #26, opened 2015-07-27
Add method extractJSON and attr "len" for TextPage

deepgully:
extract text with coordinates to JSON format.


rk700:
Great! I guess you can also make a PR to the upstream:)

PR #30, opened 2015-11-17
required changes for PyMuPDF-1.8

ousia:
I hope this is the right branch to do the following changes:

* removed LibLinux and LibWin32 directories,

* removed PyMuPDF.chm.zip and PyMuPDF.pdf.zip,

* expanded PyMuPDF.pdf,

* added new submodule for PyMuPDF-optional-material.

Tell me whether you need these changes in master.


rk700:
LGTM. 

Then the last step would be to update `library_dirs` in setup.py, and put the command details in the README since windows users should use `git submodule` to pull the 3rd party libs.

PR #57, opened 2016-06-04
Move `mupdfthird` to the end to avoid over link for Linux

mozbugbox:
Any shared libraries from the system that's desired should be put in
front of `mupdfthird`. Otherwise, both the static and shared libs of the
same lib will be linked into _fitz.so.


rk700:
Great. Thank!.

PR #119, opened 2018-01-06
New methods for vector image input & output

JorjMcKie:


JorjMcKie:
Extend v1.12.1 with vector image capabilities

PR #190, opened 2018-07-24
Fixes in an Example script and added shell scripts for installation in CentOS and Ubuntu

asshatter:
Minor fix in TableExtract script and added shell scripts for CentOS and Ubuntu. Might need to change url when a new version is released.


JorjMcKie:
Thanks a lot!

PR #273, opened 2019-03-30
Update ubuntu_pymupdf.sh

wenheping:
Current PyMuPDF should build with mupdf-1.14.0
Add some prerequirement when build under Ubuntu OS


JorjMcKie:
Thank you for your contribution!

PR #275, opened 2019-04-01
Fix the path

wenheping:
Fix the path of fitz/config.h


JorjMcKie:
Thanks for catching this!

PR #277, opened 2019-04-03
create installation for FreeBSD

wenheping:
create installation for FreeBSD


JorjMcKie:
Thanks for the contribution!
Let me ask for some more modifications:

* The repository is now part of an organization (with the same name ``pymupdf``). So every reference to it should be `/pymupdf/PyMuPDF` instead of `/rk700/PyMuPDF`.

* I believe that downloads from MuPDF website must always be `https://mupdf.com/downloads/archive/...` to be successful.

Thank you.

wenheping:
You are right, now fixed.

wen
________________________________
发件人: Jorj X. McKie <notifications@github.com>
发送时间: 2019年4月3日 23:20
收件人: pymupdf/PyMuPDF
抄送: wenheping; Author
主题: Re: [pymupdf/PyMuPDF] create installation for FreeBSD (#277)


Thanks for the contribution!
Let me ask for some more modifications:

  *   The repository is now part of an organization (with the same name pymupdf). So every reference to it should be /pymupdf/PyMuPDF instead of /rk700/PyMuPDF.

  *   I believe that downloads from MuPDF website must always be https://mupdf.com/downloads/archive/... to be successful.

Thank you.

―

Issue #370, opened 2019-10-01
The Example for Embedded Import doesn't work. 


jamesacampbell:
_**Please provide all mandatory information!**_

## Describe the bug (mandatory)
I am trying to use the method `embeddedFileAdd` from `fitz` via the example file referenced below `embedded-import.py`.

## To Reproduce (mandatory)

Run the example code at https://github.com/pymupdf/PyMuPDF/blob/master/examples/embedded-import.py and you will see the error is the following:

```
  File "/usr/local/lib/python3.7/site-packages/fitz/fitz.py", line 2374, in _embeddedFileAdd
    return _fitz.Document__embeddedFileAdd(self, name, buffer, filename, ufilename, desc)
TypeError: in method 'Document__embeddedFileAdd', argument 2 of type 'char const *'
```
This appears to be due to incorrect handling of unicode for C bindings inside of fitz, according to [this](https://stackoverflow.com/questions/20905702/typeerror-argument-of-type-char-const) stack overflow answer.

## Expected behavior (optional)
It should successfully embed an attachment into the pdf and output a modified pdf.

## Your configuration (mandatory)
 - Mac OS Mojave
 - Python 3.7.3
 - PyMuPDF version via pip3 install https://files.pythonhosted.org/packages/d2/23/561914a111f2990df21563385e377e274a4f01f178b52953e48be0c4a8ff/PyMuPDF-1.16.3-cp37-cp37m-macosx_10_6_intel.whl

For example, the output of `print(sys.version, "\n", sys.platform, "\n", fitz.__doc__)` would be sufficient (for the first two bullets).

jamesacampbell:
Note, I tried two different PDFs as the source files to ensure that was not the issue.

jamesacampbell:
I also tried different attachments.

JorjMcKie:
Please referto the **_current_** documentation:
First and second argument have been reversed two or so versions ago.
Please modify the example accordingly. You can also submit a PR once you can confirm it is working.

jamesacampbell:
Wow @JorjMcKie it worked fine as soon as I flipped it. Thanks! I will submit a PR for the example file. I can confirm it works and can see the embedded file in PDF Expert app as an attachment (I don't see a way to see it in Preview.app, but that is a limitation in Preview I believe).

jamesacampbell:
Correct me if I am wrong, but I believe this is the only Python PDF package that can embed attachments?

JorjMcKie:
[...]

PR #371, opened 2019-10-01
Update embedded-import.py

jamesacampbell:
Fixed method call as described in the original https://github.com/pymupdf/PyMuPDF/issues/370


JorjMcKie:
Thank you!

PR #449, opened 2020-02-23
Prevent docs .ico build warning

bgermann:
A system with a case sensitive file system will show:
WARNING: favicon file 'pymupdf.ico' does not exist


JorjMcKie:
Thank you very much for catching and correcting this!

PR #486, opened 2020-04-08
typo in def addHighlightAnnot

mindstorm75:
quads = get_highlight_selection(self, start=start, stop=stop, clip=rect)
should be
quads = get_highlight_selection(self, start=start, stop=stop, clip=clip)


JorjMcKie:
Oops - thanks a lot!

Issue #574, opened 2020-07-30
Please fix the linking for openSUSE


jloehel:
## Describe the bug (mandatory)
The linking does not really work for openSUSE.

## To Reproduce (mandatory)
```
╭─jloehel@workmachine /tmp/test_pymupdf   
╰─➤  python3 -m venv .venv && source .venv/bin/activate                                                                                        
╭─jloehel@workmachine /tmp/test_pymupdf  (.venv) 
╰─➤  pip install pymupdf 
Collecting pymupdf
Installing collected packages: pymupdf
Successfully installed pymupdf-1.17.4
You are using pip version 18.1, however version 20.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
╭─jloehel@workmachine /tmp/test_pymupdf  (.venv) 
╰─➤  python
Python 3.6.10 (default, Jan 16 2020, 09:12:04) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import fitz
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/test_pymupdf/.venv/lib64/python3.6/site-packages/fitz/__init__.py", line 3, in <module>
    from fitz.fitz import *
  File "/tmp/test_pymupdf/.venv/lib64/python3.6/site-packages/fitz/fitz.py", line 18, in <module>
    from . import _fitz
ImportError: /tmp/test_pymupdf/.venv/lib64/python3.6/site-packages/fitz/_fitz.cpython-36m-x86_64-linux-gnu.so: undefined symbol: jpeg_resync_to_restart
>>> exit()
```

## Expected behavior (optional)
The installation from pip works without patching. (https://github.com/jloehel/PyMuPDF/commit/22e525391be67fd98ae72db87f33b03d70535296)

## Your configuration (mandatory)
```
╭─jloehel@workmachine /tmp/test_pymupdf  (.venv) 
╰─➤  cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.1"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.1"
PRETTY_NAME="openSUSE Leap 15.1"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.1"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
[...]

PR #579, opened 2020-08-02
Load libraries on Fedora

cges30901:
This pull request is related to https://github.com/pymupdf/PyMuPDF/issues/574. Based on https://github.com/pymupdf/PyMuPDF/pull/577, this commit also detects Fedora and loads needed libraries. Tested on Fedora 32.

I also want to add Ubuntu support, but since Travis uses Ubuntu to build, I am not sure if linking additional libraries will affect builds. I am not familiar with Travis, maybe setup.py also needs to detect travis?


JorjMcKie:
> I also want to add Ubuntu support, but since Travis uses Ubuntu to build, I am not sure if linking additional libraries will affect builds. I am not familiar with Travis, maybe setup.py also needs to detect travis?

Ubuntu is indeed used by Travis, and Travis uses `setup.py` provided in the package. Therefore Ubuntu **_is_** already covered by the current setup script.

PR #591, opened 2020-08-14
Load libraries on Manjaro

cges30901:


JorjMcKie:
Thanks a lot for your efforts!

PR #626, opened 2020-08-27
Build on any GNU system

pinotree:
The toolchain of GNU systems mostly is the same as the ones used on Linux, so share the build settings.


JorjMcKie:
Thank you!

PR #627, opened 2020-08-27
fix some spelling errors

cbm755:
Here's a few spelling errors I've fixed while looking through the docs.

I was methodical in one file (faq.rst I think) but others are just random things I noticed.

Should be uncontroversial except `gluing`: I swear `glueing` looks right but dictionary says its `gluing`!


JorjMcKie:
Thank you very much indeed 😎

PR #682, opened 2020-10-09
build against system gumbo library on Fedora

mjg:
On Fedora, we build mupdf against as many system libraries as possible (all but the ones patched by Artifex). This includes the gumbo parser lib.

This is not the only change which PyMuPDF 1.18.0 requires: Since 1.17.5 PyMuPDF requires to be built against patched mupdf sources which really is a no-go.

For now, I'm including the patch for jm_valid_chars() into mupdf packages in Fedora so that - together with this PR for PyMuPDF - the package PyMuPDF can stay in Fedora.

But please do not make this into a habit. It creates packaging nightmares, and mupdf is developping its own python bindings in the 1.18 series. So I suggest upstreaming jm_valid_chars() (that snippet is used in mupdf anyways, just not as a separate function yet) or using the new binding.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ class build_ext_first(build_py_orig):
 DEFAULT = ["mupdf", "mupdf-third"]
 ARCH_LINUX = DEFAULT + ["jbig2dec", "openjp2", "jpeg", "freetype"]
 OPENSUSE = ARCH_LINUX + ["harfbuzz", "png16"]
-FEDORA = ARCH_LINUX + ["harfbuzz"]
+FEDORA = ARCH_LINUX + ["harfbuzz", "gumbo"]
 LIBRARIES = {
     "default": DEFAULT,
     "arch": ARCH_LINUX,


JorjMcKie:
> But please do not make this into a habit. 

Understood. I will see to find another solution. This was one was out of laziness to be honst, because otherwise I would have needed to include more include libraries in the PyMuPDF builds. But this is perfectly possible, and I'll go this way.

Apart from this "luxury" case, the more general issue however is this: timely reaction to serious MuPDF bugs occasionally requires changing MuPDF code within their **current** version - before a new MuPDF is out.
This was the case in the later 1.17 patches - for a time period of 6 months or so.
I am also providing my own config.h which has a hefty impact on the PyMuPDF binary.
I hear you saying this shouldn't happen either anymore ...?

I was also thinking of making my MuPDF fork that would incorporate any of my fixes. That wouldn't help you either or would it?

mjg:
> Understood. I will see to find another solution. This was one was out of laziness to be honst, because otherwise I would have needed to include more include libraries in the PyMuPDF builds. But this is perfectly possible, and I'll go this way.

mupdf itself is no good example, they think forking and bundling is the way to go. But maybe they're open to breaking out the functions that you need. (Artifex tried to work with lcms2 upstream and failed, which justifies the fork, but not bundling everything else.)

> Apart from this "luxury" case, the more general issue however is this: timely reaction to serious MuPDF bugs occasionally requires changing MuPDF code within their **current** version - before a new MuPDF is out.
> This was the case in the later 1.17 patches - for a time period of 6 months or so.

Yes, that indeed is a problem - also because of the missing ABI compatibility in jbig2dec which mupdf and ghostscript depends on. We try to coordinate builds now.

> I am also providing my own config.h which has a hefty impact on the PyMuPDF binary.
> I hear you saying this shouldn't happen either anymore ...?

I'm not the PyMuPDF main packager in Fedora, but afaics it was always built against the system mupdf without your "underscore patches", which - from the Fedora packaging perspective - was the right thing to do. Fedora's mupdf builds with these flags: `-fPIC -DJBIG_NO_MEMENTO -DTOFU -DTOFU_CJK` It may overlap with some of your config changes.

This broke in 1.17.5 when the underscore patches introduced a new exported symbol (function) to mupdf.

> I was also thinking of making my own MuPDF fork that would incorporate any of my fixes. That wouldn't help you either or would it?

No, that would make things worse since we have to follow the official upstream unless it is dead.Have you looked at the new python bindings? I have not packaged them yet but I would hope that they help and - since they are experimental - you have a chance to help shape them the way you need them.

JorjMcKie:
Ok, thanks for detailing this out.

As for the new MuPDF Python bindings:
I had a rough look into it. As for the general direction, they do much the same thing as I am doing - sort of a competition if you will, also using SWIG, etc.
They are very early in this development, but it so far looks like a replication of what they did with their JavaScript bindings: very low level, very challenging for an **_application_** programmer. You have little chance without knowing a lot about MuPDF's technical architecture: for every MuPDF C function, there is a Python method - I hope you see what I mean.
All that is - I hope - very different in PyMuPDF.
But you are right: I should offer my collaboration and see how they react.

mjg:
> Ok, thanks for detailing this out.
> 
> As for the new MuPDF Python bindings:
> I had a rough look into it. As for the general direction, they do much the same thing as I am doing - sort of a competition if you will, also using SWIG, etc.
> They are very early in this development, but it so far looks like a replication of what they did with their JavaScript bindings: very low level, very challenging for an **_application_** programmer. You have little chance without knowing a lot about MuPDF's technical architecture: for every MuPDF C function, there is a Python method - I hope you see what I mean.
> All that is - I hope - very different in PyMuPDF.
> But you are right: I should offer my collaboration and see how they react.

Oh yes, I see where they are going. But maybe upstream could provide the low level bindings that you need to provide the pythonish high level access, hoping those are easier to follow (in terms of updates) since the C api. Otherwise there is not much gain, obviously.

PR #694, opened 2020-10-21
Don't decode streams in utils.scrub

DavideWalder:
In `utils.scrub`, when removing hidden text, there is an attempt to UTF-8 decode the `\Contents` stream, which is not necessarily possible since it also contains text strings which according to the official reference guide can be arbitrary byte arrays.

The proposed changes address the issue by comparing the byte arrays without decoding.

Quote from section 5.3.2 of the [official reference guide](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdf_reference_archive/pdf_reference_1-7.pdf):

_The strings must conform to the syntax for string objects. When a string is written by enclosing the data in parentheses, bytes whose values are the same as those of the ASCII characters left parenthesis (40), right parenthesis (41), and backslash (92) must be preceded by a backslash character. **All other byte values between 0 and 255 may be used in a string object.** These rules apply to each individual byte in a string object, whether the string is interpreted by the text-showing operators as single-byte or multiple-byte character codes._

### DIFF
--- a/fitz/utils.py
+++ b/fitz/utils.py
@@ -3467,23 +3467,23 @@ def scrub(
         suppress = False  # indicate text suppression active
         make_return = False
         for line in cont_lines:
-            if line == "BT":  # start of text object
+            if line == b"BT":  # start of text object
                 in_text = True  # switch on
                 out_lines.append(line)  # output it
                 continue
-            if line == "ET":  # end of text object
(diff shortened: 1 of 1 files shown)


JorjMcKie:
Thank you very much for your contribution!

PR #717, opened 2020-11-11
fix incorrect func calls witihin `Pixmap().pillowData()`

tjcuddihy:
Trying to use `pillowData("JPEG")` and I got some errors. The proposed changes allow for expected usage of `pillowData()`


JorjMcKie (approved):

JorjMcKie:
Thank you!

PR #825, opened 2021-01-12
Fix return value order in appendix 2

user202729:
The correct argument order is shown in [the source code](https://github.com/pymupdf/PyMuPDF/blob/master/fitz/fitz.i#L9831) and [a documentation file](https://github.com/pymupdf/PyMuPDF/blob/master/docs/textpage.rst).


JorjMcKie:
Thank you for spotting this!

PR #832, opened 2021-01-13
Several small changes

user202729:
(I didn't test the `extractTEXT` change, but it looks correct. There wasn't `extractTEXT` before, despite it being documented.)


JorjMcKie (approved):
Thank you very much indeed!

PR #835, opened 2021-01-16
Add snake-case alias for many methods

user202729:
It's really hard to find out the correct place to put those methods (sometimes putting in `__init__.py` is necessary because the method is assigned there, but currently most of the time it's assigned in the class definition; however that method doesn't work for properties; defining an alternative property that returns the original works but doesn't look very good)

(The list is semi-autogenerated.)

It would be much harder to modify the documentation automatically, however...

Besides, some of those are actually redundant (because `IRect` inherits from `Rect`, for example) -- on the other hand, there are some existing redundant assignments already.


JorjMcKie:
Thank you for this!
I am sure you have seen, that I am underway with the exact same intention.
For new stuff I am already following that naming convention.

As sort of a general approach, I tried to change the old method definition to snake case and define an alias for the old name. I hope this would make it easier one day to get rid of all the old names by just not defining them.

Having the zillions of aliases also bloats the code completion / IntelliCode lists - so maybe a function would make sense that removes those aliases, when the programmer does not need them ...

JorjMcKie:
Meanwhile I have decided on an overall approach on this, which supersedes your PR:
1. Class by class I will change method and attribute names according to the *snake_case* convention.
2. For each name change, I put an alias definition ``class.oldName = class.new_name`` statement in a function ``restore_aliases`` defined in ``__init__.py``.
3. For the time being, function ``restore_aliases`` is being executed on importing PyMuPDF.
4. In one of the next versions, maybe 1.19.0, the automatic execution of the alias function will be dropped, but remain available for those who need backward compatibility in existing scripts - they just need to execute `fitz.restore_aliases()` after the import.

PR #870, opened 2021-01-29
allow conversion from PDF to PDF

veecue:
`mutool convert` allows PDF as input and output type. There is no sense
in forbidding this action in the python library. A possible use case
for converting a PDF to a PDF using mupdf is rendering PDF form widgets
as PDF text.

### DIFF
--- a/docs/document.rst
+++ b/docs/document.rst
@@ -625,7 +625,7 @@ For details on **embedded files** refer to Appendix 3.
 
     .. method:: convertToPDF(from_page=-1, to_page=-1, rotate=0)
 
-      Create a PDF version of the current document and write it to memory. **All document types** (except PDF) are supported. The parameters have the same meaning as in :meth:`insert_pdf`. In essence, you can restrict the conversion to a page subset, specify page rotation, and revert page sequence.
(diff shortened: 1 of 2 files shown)


JorjMcKie:
Thanks for your contribution! Some background on my motivation to exclude this initially:
It is a functional (not technical) subset of what has ever been available as `doc.write(..)` (or the new name `doc.tobytes()`). But it won't do any harm either of course.

veecue:
Turns out that aparrently it's not even a functional subset. At least for my use case, I couldn't get it working using the write() method.

JorjMcKie:
How do you mean: you created a `Document` with the resulting bytes and it wouldn't work?

veecue:
What I tried to do: Turning (rendering) a Document containing a form to a document containing text. I needed that because the printer wouldn't print the form properly. I couldn't achieve that using the `save()` method and any options I passed to it. However using `convertToPDF()` after this patch or using `mutool convert` on the commandline, it did just what I wanted.

JorjMcKie:
When I use it, I get a completely re-created PDF which looks alike, but has a rejuvenated internal structure: an existing PDF-1.x is turned into the MuPDF standard PDF-1.7, etc., etc.

JorjMcKie:
Ah, I finally got you. Okay then.

JorjMcKie:
To be honest, I wasn't even aware of this MuPDF convert feature ... and would never have expected that!

PR #928, opened 2021-03-02
fix invert matrix bug

zhou3968322:
invert matrix cause memory leak

### DIFF
--- a/fitz/fitz.i
+++ b/fitz/fitz.i
@@ -11089,7 +11089,11 @@ struct Tools
                 a = -src.e * dst.a - src.f * dst.c;
                 dst.f = -src.e * dst.b - src.f * dst.d;
                 dst.e = a;
-                return Py_BuildValue("(i, O)", 0, JM_py_from_matrix(dst));
+                PyObject* mat = JM_py_from_matrix(dst);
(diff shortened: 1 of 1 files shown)


JorjMcKie:
This was unnoticed for a long time - thank you!

zhou3968322:
I feel that debugging pymupdf in Mac is a bit troublesome. &nbsp;Do you have any suggestions about how to debug Swig?



发自我的iPhone


------------------ Original ------------------

JorjMcKie:
I don't debug SWIG itself ever actually. Instead I either debug the .py or .c files it generates.
After introducing new functions, I routinely perform memory leak tests focussed on them (using GC). The strategy here is generally executing the new feature many thousands of times, freeing memory between each and compare before and after memory occupation status.
> The one you detected went unnoticed because of its small memory amount. BTW the leak can also be prevented simply by using string "iN" instead of "iO" in `Py_BuildValue`. Embarrassingly, the Python documentation explicitely recommends this in cases as mine 😒.

For **_Python_** debugging, I found VS Code (my editor) utterly sufficient.
If a have a problem in some **_C_** part, I use a Python script to automatically insert tracing statements for the failing part - not a C debugger.

PR #988, opened 2021-03-31
Add new homebrew directories to build script

b4stien:
These added directories ensure that PyMuPDF is installable on an ARM M1 Mac (after having run at least `brew install mupdf swig`).

For reference Homebrew switch for `/opt/homebrew` as a base directory for "Apple Silicon" installations, cf https://docs.brew.sh/Installation.


b4stien:
You're welcome, I've encountered the same struggle than https://github.com/pymupdf/PyMuPDF/discussions/875, it'll at least allow us to build/install locally while the wheel is getting fixed :)

Thank *you* for your work on this package!

PR #999, opened 2021-04-09
new saveSane() method as shortcut for sane defaults

fgallaire:


JorjMcKie:
Thank you very much for the contribution! I appreciate your wish to help.
In your case however, I was literally minutes before publishing a new version (1.18.12), so I quickly included "your" new save method before seeing you did the same.
I gave it the name ``ez_save`` ("easy save"). It does have the defaults you liked to see: ``garbage=3, deflate=True``.
Thanks again!

fgallaire:
@JorjMcKie ok thanks !!

fgallaire:
@JorjMcKie thanks a lot your version is a lot better, with sane defaults but still completely configurable, really nice !

PR #1133, opened 2021-07-09
Update font.rst

inf3rnus:
I think the argument ``filename`` was from a previous version of the software for v1.18.14 it is no longer a kwarg, so it should be ``fontfile`` if I am not mistaken!


JorjMcKie (approved):

JorjMcKie:
You are right - thank you!

PR #1182, opened 2021-07-31
fix broken link layout preserving text extraction

danofsteel32:
[textlayout.py](https://github.com/pymupdf/PyMuPDF-Utilities/blob/master/text-extraction/textlayout.py) no longer points to anything. That functionality appears to be provided by [fitzcli.py](https://github.com/pymupdf/PyMuPDF-Utilities/blob/master/text-extraction/fitzcli.py) via it's `gettext` option now. Also propose linking to the documentation on using the `gettext` option [here](https://github.com/pymupdf/PyMuPDF-Utilities/tree/master/text-extraction#layout-preserving-text-extraction).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,9 @@ Have a look at the basic [demos](https://github.com/pymupdf/PyMuPDF-Utilities/tr
 
 **_New: Layout preserving text extraction!_**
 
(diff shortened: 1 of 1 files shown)


JorjMcKie:
Thank you - you are absolutely right.
This `fitzcli.py` is of interest only as long as the new v1.18.16 is not out. After that, as mentioned, `gettext` is available in the `fitz` module.
I replaced the previous script by `fitzcli.py` in a hurry to prevent people from getting used to stuff that would soon be obsolete.

PR #1221, opened 2021-08-21
add a test: extract images

ph3n92h3:
Hello!😊

Today I used PyMuPDF when I did some exercises about automation for pdf file. I wrote a program to extract images from a pdf file, and I think it would be a proper test, even it can be used during daily work. So I submit this application.

Good luck!🤗
[test_extract_image.zip](https://github.com/pymupdf/PyMuPDF/files/7025701/test_extract_image.zip)

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -23,6 +23,7 @@ The following areas are currently covered:
 * handling of PDF Tables of Contents
 * annotation handling
 * field / widget handling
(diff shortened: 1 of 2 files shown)


JorjMcKie:
Thank you very much for your contribution! This type of thing is what should happen a lot more often!
I will accept it for now and look at details later.

PR #1225, opened 2021-08-22
Deduplicate image extraction test

mara004:
See my two comments on the PR:
https://github.com/pymupdf/PyMuPDF/commit/7d7aa3a11b1268983d80067b5f618e2a783f0697#r55240385
https://github.com/pymupdf/PyMuPDF/commit/bfd91fd8c92dad59b40f9d2c7dbd5d410b8bb763#r55240444

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -23,7 +23,7 @@ The following areas are currently covered:
 * handling of PDF Tables of Contents
 * annotation handling
 * field / widget handling
-* extraction images
+* image extraction
 
 This is **_not a coverage test_**, although a significant part of the relevant Python part **_does_** get executed (ca. 80%). Achieving a much higher code coverage remains an ongoing task.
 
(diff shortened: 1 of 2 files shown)


JorjMcKie:
In the meantime, I ahve have published a new version, 1.18.17.
It includes a test script for image extraction, that very roughly is based on your contribution. I take the opportunity for some comments on scripts in the ``tests`` folder:
1. They should **_either_** test something that can be checked against some **_expected result_** (like something that can be verified with the ``assert`` statement), **_or_** simply make sure to extend the number of lines executed as such (coverage extension).
2. Formally, only functions with names starting with ``test_`` or ending with ``_test`` are executed by ``pytest``. Everything else is ignored. The execution happens based on importing scripts that follow the same naming convention.

So the new test script extracts the images from a PDF, counts each extraction and compares the total count with a fixed number. It **_does not store_** the image to anywhere, because that would not test any PyMuPDF functionality. This (1) reduces the code drastically and (2) shortens execution speed.
When scanning the PDF xrefs, I also skip non-image objects. This prevents ugly MuPDF error messages.

PR #1294, opened 2021-09-27
Use system tmp folder for build_subset artifacts

### LINKED ISSUE #1293: Support temporary directory in `subset_fonts`
**Is your feature request related to a problem? Please describe.**
[subset_fonts](https://pymupdf.readthedocs.io/en/latest/document.html#Document.subset_fonts) creates temporary files in the current directory https://github.com/pymupdf/PyMuPDF/blob/master/fitz/utils.py#L4977  which is a problem when the directory is not writable. For example, if used in AWS Lambda, only `tmp` is writable.

**Describe the solution you'd like**
Use system's temp folder https://docs.python.org/3/library/tempfile.html

**Describe alternatives you've considered**
Path to tmp dir parameter

manycoding:
Fixed #1293

### DIFF
--- a/fitz/utils.py
+++ b/fitz/utils.py
@@ -705,8 +705,6 @@ def get_pixmap(page: Page, **kw) -> Pixmap:
     pix = dl.get_pixmap(matrix=matrix, colorspace=colorspace, alpha=alpha, clip=clip)
     dl = None
     return pix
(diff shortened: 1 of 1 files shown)


JorjMcKie:
Thanks a lot!

PR #1356, opened 2021-10-30
Reorder Linux linking and introduce DEBIAN

bgermann:
harfbuzz is in the shared libraries list for some distros.
Let them depend on each other and add Debian.


JorjMcKie:
Thank you!

PR #1363, opened 2021-11-02
update page for winding rules in the manual

mlissner:
Looks like the old manual is gone and the new one has it on [page 136](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=136)


JorjMcKie:
Thank you!

PR #1421, opened 2021-11-24
Fix Example

Luux:
Small fix for one the example snippets in the FAQ.

`draw_rect` results in `ValueError: bad operand type`, so we should use `draw_quad` instead.


JorjMcKie:
Thank you 👍

PR #1530, opened 2022-01-12
test was leaving a file behind

cbm755:
I put it in a temporary directory using pytest's "tmpdir" feature.

It didn't look like it was reading it back so I read it back and did a
(weak?) comparison (based on reading `test_pilsave`).

### DIFF
--- a/tests/test_pixmap.py
+++ b/tests/test_pixmap.py
@@ -64,17 +64,15 @@ def test_pilsave():
         pass
 
 
-def test_save():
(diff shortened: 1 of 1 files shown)


cbm755:
Forgot to say, without this I always got a "11" or "17" or some such png file in my test directory.

JorjMcKie:
Great, thank you!

JorjMcKie:
Interestingly, this doe not happen on Windows.

PR #1731, opened 2022-05-27
Adding correct link for shapes_and_symbols.py

qwertynik:


jamie-lemon:
Hello - thanks for this - the file you reference should probably be updated to: https://github.com/pymupdf/PyMuPDF-Utilities/blob/master/shapes/shapes_and_symbols.py

PR #1823, opened 2022-07-19
Clarify possible return codes of get_sigflags()

JorjMcKie:
Recommend using return code bit testing instead of integer values.


julian-smith-artifex-com:
These changes were merged in a different pull request.

PR #1903, opened 2022-08-30
Fix typos on documentation

10ego:
method page.getTex() does not exist in the example.

use of the `:meth:` sphinx syntax produces a paranthesis at the end, rendering `:meth:Page.get_text("words")` like **`Page.get_text("words")()`**.

Consider swapping `meth` with `const` to keep consistent rendering format, or remove the value passed into the method.

### DIFF
--- a/docs/recipes-text.rst
+++ b/docs/recipes-text.rst
@@ -109,7 +109,7 @@ But you also have other options::
      """Underline each word that contains 'text'.
      """
      found = 0
-     wlist = page.getTex("words")  # make the word list
+     wlist = page.get_text("words")  # make the word list
      for w in wlist:  # scan through all words on page
          if text in w[4]:  # w[4] is the word's string
              found += 1  # count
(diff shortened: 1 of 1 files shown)


julian-smith-artifex-com:
According to https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-method `:const:` is _Reference a “defined” constant. This may be a Python variable that is not intended to be changed._, which doesn't seem right.

So rather than:

    This script uses :const:`Page.get_text("words")` to look for a string

perhaps we should just do:

    This script uses ``Page.get_text("words")`` to look for a string

?

10ego:
That probably works better.  Updated the branch with a new commit.

cgdae:
Thanks for this. Could you squash the two commits into one? Then it should be ok to merge it.

cgdae:
The commit message "Update method highlighting with an example input value" seems wrong - could you change back to "Fix typos on documentation"?

10ego:
commit message amended as suggested

julian-smith-artifex-com:
Thanks, now merged.

PR #1904, opened 2022-09-01
Correct default dpi

serereg:


JorjMcKie:
This change should be rejected. The current documentation correctly states that the default x- and y-resolution of a page Pixmap are 96, and not 72.
Can easily be confirmed by saving a PNG from the Pixmap and checking the stored DPI via some tool, e.g. Irfanview: it will be 96 x 96!

jamie-lemon:
Indeed, just checked the output and can validate it is indeed @96 dpi

julian-smith-artifex-com:
It might not matter too much now, but did the confusion here come from mutool draw having a default resolution of 72? (see https://git.ghostscript.com/?p=mupdf.git;a=blob;f=source/tools/mudraw.c#l306)

PR #1942, opened 2022-10-03
Fix letter generation for page labels

FelixFrog:
The old code generates page labels like a,b,c,d,...,x,y,z,aa,bb,cc,... while it should be a,b,c,d...x,y,z,aa,ab,ac,... This code should also work for a number way bigger than what any reasonable individual would need for a pdf page label


JorjMcKie:
Thank you!

PR #1944, opened 2022-10-03
Fixes for leaks in test suite.

julian-smith-artifex-com:


JorjMcKie (approved):
As mentioned in Notion yesterday, I will try to unify the two functions JM_get_fontextension() and JM_get_fontbuffer(). I once copied one from the other a long time ago when I was in a hurry and didn't want to spend the effort of extending the older one.
But this is not urgent I believe, so we can go as it is now.

Issue #1956, opened 2022-10-10
Skipping link: none of the wheel's tags are compatible


programarivm:
Hi there,

I'm still trying to install PyMuPDF after successfully installing the `python3-dev` and `libfreetype6-dev` packages as described in https://github.com/pymupdf/PyMuPDF/issues/1955. Now I've just logged the installation output. 

```
$ pip install pymupdf --log log.txt
```

And obtained the following `log.txt`:

```
2022-10-10T12:32:53,752 Using pip 21.3.1 from /usr/local/lib/python3.6/dist-packages/pip (python 3.6)
2022-10-10T12:32:53,754 Defaulting to user installation because normal site-packages is not writeable
2022-10-10T12:32:53,868 Created temporary directory: /tmp/pip-ephem-wheel-cache-mqiawlww
2022-10-10T12:32:53,868 Created temporary directory: /tmp/pip-req-tracker-8bvblrht
2022-10-10T12:32:53,868 Initialized build tracking at /tmp/pip-req-tracker-8bvblrht
2022-10-10T12:32:53,868 Created build tracker: /tmp/pip-req-tracker-8bvblrht
2022-10-10T12:32:53,868 Entered build tracker: /tmp/pip-req-tracker-8bvblrht
2022-10-10T12:32:53,868 Created temporary directory: /tmp/pip-install-6p3lbjfc
2022-10-10T12:32:53,889 1 location(s) to search for versions of pymupdf:
2022-10-10T12:32:53,889 * https://pypi.org/simple/pymupdf/
2022-10-10T12:32:53,889 Fetching project page and analyzing links: https://pypi.org/simple/pymupdf/
2022-10-10T12:32:53,890 Getting page https://pypi.org/simple/pymupdf/
2022-10-10T12:32:53,891 Found index url https://pypi.org/simple
2022-10-10T12:32:54,310   Skipping link: none of the wheel's tags (cp27-cp27m-win32) are compatible (run pip debug --verbose to show compatible tags): https://files.pythonhosted.org/packages/4f/af/876883fc5e06a81dc9c1120fb2819bc301ad74d6134e544ff92daa4c0e31/PyMuPDF-1.11.2-cp27-cp27m-win32.whl#sha256=aa1c7412784c044819d8e51f891f765e5538ff85f64715c7cfa4c32435c4137e (from https://pypi.org/simple/pymupdf/)
[...]

Issue #95, opened 2017-06-22
ImportError: No module named _fitz


mpvillafranca:
Hi,

just installed PyMuPDF on my Linux Mint inside a virtualenv following the Ubuntu instructions. Everything was looking good until I called the "import fitz", geting this error:

```
>>> import fitz
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "fitz/__init__.py", line 2, in <module>
    from fitz.fitz import *
  File "fitz/fitz.py", line 17, in <module>
    _fitz = swig_import_helper()
  File "fitz/fitz.py", line 16, in swig_import_helper
    return importlib.import_module('_fitz')
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named _fitz
```
Do you know what could be wrong? I didn't get any kind of error during the building process so I suppose everything went fine on that stage...

Hope you can help me.

JorjMcKie:
I am no Linux guy - so @rk700 will need to step in.

But in the meantime can you find out, where in fact the generated _fitz.so has landed? It must exist somewhere, given you setup went correctly.

Are you sure you invoke the interpreter in the exact same environment, where PyMuPDF was generated?

A setup.py protocol would be helpful, too ...

Thanks!

rk700:
Hi,

I guess you are running python in PyMuPDF's root dir @mpvillafranca . If so, please `cd` into some other directory and try again.

mpvillafranca:
Hi,

I followed the Ubuntu steps for the installation. Here is my `setup.py`:

```
from distutils.core import setup, Extension
import sys, os

# check the platform
if sys.platform.startswith('linux'):
    module = Extension('fitz._fitz', # name of the module
                       ['fitz/fitz_wrap.c'], # C source file
                       include_dirs=['/usr/local/include/mupdf',    # we need the path of the MuPDF and zlib$
                                     '/usr/local/thirdparty/zlib',  # '/usr/include/mupdf',
                                    ],
                       #library_dirs=['<mupdf_and_3rd_party_libraries_dir>'],
                       libraries=[
                           'mupdf',
                           #'crypto', #openssl is required by mupdf on archlinux
                           #'jbig2dec', 'openjp2', 'jpeg', 'freetype',
                           'mupdfthird',
                           ], # the libraries to link with
                      )
elif sys.platform.startswith('darwin'):
    module = Extension('fitz._fitz', # name of the module
[...]

PR #1957, opened 2022-10-10
Improved the docs

programarivm:
Fixes https://github.com/pymupdf/PyMuPDF/issues/1956

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,7 +1,7 @@
 Installation
 =============
 
-PyMuPDF should be installed using pip with::
+PyMuPDF must be installed using pip as it is described next::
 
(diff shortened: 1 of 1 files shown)


julian-smith-artifex-com:
Thanks for the suggested documentation change, but i don't think it's necessary.

One can install PyMuPDF in different ways (e.g. specifying a particular version number) and end up with a working installation, so changing "should" to "must" is, i think, too strong.

programarivm:
Okay, no problem. The statement would be in line with @JorjMcKie as discussed in https://github.com/pymupdf/PyMuPDF/issues/95:

> Always follow the recommendation: `python -m pip install` ... and never just `pip install` ...!

PR #1960, opened 2022-10-11
tests/: updated tests/test_insertpdf.py:test_joining() for latest mupdf.

julian-smith-artifex-com:
Latest MuPDF has changed internals that effect internal details; appearence is fine so this commit updates the expected output to match the new internals.

Also removed old tests/resources/joined-1.20.pdf and associated testing as not useful any more.


JorjMcKie (approved):
I think this is okay for now. But based on our recent discussions I think I should come up with a more stable function, which no longer depends on binary file content but focusses on correct functioning of the insertion process as such.

PR #1970, opened 2022-10-16
First version including Story material

JorjMcKie:
This contains a lot of changes and also multiple new files - especially for the "recipe" topics. Among the recipe data are a SQLITE database and a ZIP file with HTML and images of a version for a Wikipedia article.


docs/filmfestival-sql.py
```diff
@@ -0,0 +1,113 @@
...
+template.remove()  # remove the template
+
+# -------------------------------------------------------------------
+# generate the PDF
+# -------------------------------------------------------------------
+writer = fitz.DocumentWriter(__file__.replace(".py", ".pdf"), "compress")
+mediabox = fitz.paper_rect("a4")  # use pages in ISO-A4 format
+where = mediabox + (72, 36, -36, -72)  # leave page borders
+
+done = 1  # end of output indicator
```
julian-smith-artifex-com, line 105: I think this variable should be called `more`, and the code should look like:
```
more = 1
while more:
    dev = ...
    more, filled = story.place(where)
    ....
```
[I need to change `fitz/fitz.i:place()` to call its internal return variable `more` too.]

JorjMcKie: You are right - I will update this.

docs/quickfox.py
```diff
@@ -0,0 +1,86 @@
...
+COLS = 2  # layout: 2 cols 1 row
+ROWS = 1
+TABLE = fitz.make_table(WHERE, cols=COLS, rows=ROWS)
+# fill the cells of each page in this sequence:
+CELLS = [TABLE[i][j] for i in range(ROWS) for j in range(COLS)]
+
+fileobject = io.BytesIO()  # let DocumentWriter write to memory
+writer = fitz.DocumentWriter(fileobject)  # define the writer
+
+done = 1
```
julian-smith-artifex-com, line 71: As earlier, we i think we should use `more` and the code should be:
more = 1
while more:
    dev = ...
    for cell in CELLS:
        more, _ = story.place(cell)
        story.draw(dev)
        if not more:
            break
    writer.end_page()  # finish the PDF page

docs/recipes-stories.rst
```diff
@@ -4,91 +4,214 @@
...
+    <p style="font-family: sans-serif;color: blue">Hello World!</p>
+    """
+
+    MEDIABOX = fitz.paper_rect("letter")  # output page format: Letter
+    WHERE = MEDIABOX + (36, 36, -36, -36)  # leave borders of 0.5 inches
+
+    story = fitz.Story(html=HTML)  # create story from HTML
+    writer = fitz.DocumentWriter("output.pdf")  # create the writer
+    
+    done = 1  # will indicate end of input once it is set to 0
```
julian-smith-artifex-com, line 37: Suggest renaming this to `more`.

docs/filmfestival-sql.py
julian-smith-artifex-com: Great, thanks. There are some other similar occurrences later on - would you be ok to update them also?

docs/recipes-stories.rst
JorjMcKie: done already

PR #1990, opened 2022-10-25
Update mupdf version plus minor changes.

julian-smith-artifex-com:
MuPDF master was recently updated to 1.21.0, which was breaking our daily github action "Test latest MuPDF".


JorjMcKie (changes requested):
file code-printer.py needs a change concerning the several `Story.place()` methods.
We have 3 different stories placing content on each page, but only the main one, called `story` determines end of the loop. So its `more` must be named differently, eg `more_code` which at the bottom of the loop also must be checked via `if not more_code:` to prevent an endless loop.
Similar is true in lines 213 when the story for the TOC content is checked. We could name it `more_tocitems` or something.

julian-smith-artifex-com:
> file code-printer.py needs a change concerning the several `Story.place()` methods. We have 3 different stories placing content on each page, but only the main one, called `story` determines end of the loop. So its `more` must be named differently, eg `more_code` which at the bottom of the loop also must be checked via `if not more_code:` to prevent an endless loop. Similar is true in lines 213 when the story for the TOC content is checked. We could name it `more_tocitems` or something.

Good point, hopefully addressed in my new force-pushed commits.

JorjMcKie (approved):

PR #2004, opened 2022-10-28
Change `Xml.add_link()` method

JorjMcKie:
We should support two different parameters, the HTML "href" value, which refers to the link target, and independently the text that is visible to the user.

Also document that `Xml.add_paragraph()` is a context manager.

### DIFF
--- a/.github/workflows/test_1.21.yml
+++ b/.github/workflows/test_1.21.yml
@@ -1,0 +1,48 @@
+name: Test branch 1.21.
+
+on:
+  schedule:
+    - cron: '13 5 * * *'
+  workflow_dispatch:
+
+jobs:
+
+  test_branch_121:
+    # Simple build+test on single platform using latest MuPDF from git. This is
+    # a cut-down version of `build_wheels` except that we use latest MuPDF from
+    # git and use a single platform and python version.
+    #
+    name: Test branch 1.21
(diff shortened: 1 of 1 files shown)


julian-smith-artifex-com (approved):

docs/xml-class.rst
```diff
@@ -103,17 +103,20 @@ There is no need to ever directly construct an :ref:`Xml` object: after creating
        :arg width: if provided, either an absolute (int) value, or a percentage string like "30%". A percentage value refers to the width of the specified ``where`` rectangle in :meth:`Story.place`. If this value is provided and ``height`` is omitted, the image will be included keeping its aspect ratio.
        :arg height: if provided, either an absolute (int) value, or a percentage string like "30%". A percentage value refers to the height of the specified ``where`` rectangle in :meth:`Story.place`. If this value is provided and ``width`` is omitted, the image's aspect ratio will be honored.
 
-    .. method:: add_link(link)
+    .. method:: add_link(href=None, text=None)
```
jamie-lemon, line 106: So we have: `add_link(link)` -> `add_link(href=None, text=None)`👍 . But, does the `=None` part mean that it is an optional parameter?

julian-smith-artifex-com (approved):

PR #2051, opened 2022-11-16
Missing DPI Parameter

klausmcm:
The documentation for `get_page_pixmap` states that `Page.get_pixmap()` is invoked and includes `*args, **kwargs` as function parameters. However, passing `dpi` in `get_page_pixmap` yields `TypeError: get_page_pixmap() got an unexpected keyword argument 'dpi'`. This change fixes this behavior.

### DIFF
--- a/fitz/utils.py
+++ b/fitz/utils.py
@@ -854,6 +854,7 @@ def get_pixmap(page: Page, *args, **kw) -> Pixmap:
 def get_page_pixmap(
     doc: Document,
     pno: int,
+    dpi: int,
     matrix: matrix_like = Identity,
     colorspace: Colorspace = csRGB,
     clip: rect_like = None,
@@ -866,6 +867,7 @@ def get_page_pixmap(
         Convenience function calling page.get_pixmap.
     Args:
         pno: (int) page number
+        dpi: (int) dpi
         matrix: Matrix for transformation (default: Identity).
         colorspace: (str,Colorspace) rgb, rgb, gray - case ignored, default csRGB.
         clip: (irect-like) restrict rendering to this area.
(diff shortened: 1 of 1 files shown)


klausmcm:
Adding the `dpi` works but I'm not sure how to set up this project. Running `pytest tests`gives me the general `ModuleNotFoundError: No module named 'fitz'` error. I'd be happy to run the tests if recommended setup instructions are provided.

julian-smith-artifex-com:
Thanks for this, you've identified a genuine problem with the API.

But actually i think we can make a slightly more general solution to the problem, using Python's `*` arg support to force args to be keyword-only. I have a change for this in my tree, which i'll push in a day or two.

So i don't think we'll take your PR directly, but will instead address the issue differently. Hope that's ok.

[I'm not sure why you're seeing `ModuleNotFoundError`, i'd have to see the logs.]

julian-smith-artifex-com:
I've pushed a separate change that addresses the issue in this PR. So I'll close this PR now.

Separate from that, i can't tell what is causing the error you get from pytest. Did you build and install PyMuPDF after cloning the repository? Please open a new discussion (https://github.com/pymupdf/PyMuPDF/discussions) if you can't make progress.

klausmcm:
> Did you build and install PyMuPDF after cloning the repository?

No, this is probably the issue. Can you link me to how to do this please?

klausmcm:
Perfect. `venv` is my standard approach. I was just missing `python setup.py install`. After install `swig` and `fonttools` (pip) all tests are now passing. Thanks,

julian-smith-artifex-com:
Fixed in PyMuPDF-1.21.1.

PR #2162, opened 2023-01-07
Fixed typo

arun-mani-j:
Fixed typo.
> The reverse is also true: except :attr:`Page.rect`, resp. :meth:`Page.bound` (both *reflect* when the page is rotated), all coordinates returned by methods and attributes pertain to the unrotated page.

What does resp. mean? Should it be _and_?

### DIFF
--- a/docs/page.rst
+++ b/docs/page.rst
@@ -24,7 +24,7 @@ In a nutshell, this is what you can do with PyMuPDF:
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


arun-mani-j:

jamie-lemon:
_resp._ is shorthand for _respectively_. However we will probably reword this at some point in the future as it is not obvious!

PR #2167, opened 2023-01-09
Updates requirements.txt with specific sphinx versioning.

jamie-lemon:


JorjMcKie (approved):

jamie-lemon:
Hmmm, I think I should also remove the `docutils<0.16` dependency.

PR #2174, opened 2023-01-16
1.21

julian-smith-artifex-com:


JorjMcKie (approved):
just as an aside: Py_CLEAR(x) also sets x=NULL, so no need for an extra setting this.

julian-smith-artifex-com:
> just as an aside: Py_CLEAR(x) also sets x=NULL, so no need for an extra setting this.

Thanks, updated PR uses this.

PR #2194, opened 2023-01-26
1.21

julian-smith-artifex-com:


docs/document.rst
```diff
@@ -1241,6 +1244,28 @@ For details on **embedded files** refer to Appendix 3.
 
        2. *docsrc* TOC entries **will not be copied**. It is easy however, to recover a table of contents for the resulting document. Look at the examples below and at program `PDFjoiner.py <https://github.com/pymupdf/PyMuPDF-Utilities/tree/master/examples/PDFjoiner.py>`_ in the *examples* directory: it can join PDF documents and at the same time piece together respective parts of the tables of contents.
 
+
+    .. index::
```
jamie-lemon, line 1248: I'm not really sure what `..index` does - is it for better search indexing or something?

jamie-lemon (commented):
All looks good except for my one query, if you could confirm it for me that would be great.

docs/document.rst
julian-smith-artifex-com: Hmm, i don't know either. I think i was hoping you're be able to tell me!

I think there are quite a few other examples of the same `.. index::` pattern elsewhere in docs/document.rst.

JorjMcKie:
These index entries generate cross reference item, for which you can search.
So because pymupdf happens to call PDF joining `insert_pdf`, additional index entries take care of "merge", "join", etc. also pointing to that method.

PR #2245, opened 2023-02-21
Fixes many links to the PyMuPDF-Utilities repo scripts.

jamie-lemon:
- A recent reorganization & naming of these scripts on `PyMuPDF-Utilities` caused the links on the documentation here to
break.


julian-smith-artifex-com (approved):

JorjMcKie (commented):
Fine with me.

PR #2286, opened 2023-03-14
1.21

julian-smith-artifex-com:


JorjMcKie (approved):
If picking an arbitrary rectangle outside the redaction, which contains black (text) areas and comparing the samples of the corresponding sub-pixmap before and after applying the redaction we should now (after the fix) find equality, because the bug turned the black text to white.
This is probably a faster way to assert success.

Issue #2300, opened 2023-03-23
Too much recursion in tree (parents), makes program terminate 


Nurskss:
I encounted a strange issue, i have a pdf file that is something wrong with (looks fine) but it raises RunTimeError, too much recursion in tree (parents), When i apply i try/except clause the program just terminates instead. The error happens when i iterate over a documents pages and then try to interact with the page.

When i use page.rect the program terminates no matter what i do to avoid it, but when i use other functions like page.get_pixmap() or page.get_text('blocks'), the same issue happens but it does not end the program. Is this a bug and is there a a way to check recursion depth of three before i do anything with it?

My current solution is to just run page.get_text('blocks) and save it to a variable, and if that works then the page.rect is safe to run, but it seems a bit strange.

Nurskss:
Here is the entire code for the files, it looks a bit f** up some places because i have been trying to debug it for ages:

`import fitz
import io
from PIL import Image
from sewar.full_ref import mse
import json
import numpy as np
import re
from operator import itemgetter
from statistics import mean
from django.db import transaction
from tqdm import tqdm
from collections import Counter
import tracemalloc

#from ..engine.kartverket import GAPI
#from main.models import *
import pdb


class Takst(object):
    def __init__(self, filepath='', doc=''):
        if filepath != 'unactive':
            self.doc = fitz.open(filepath)
        if doc:
            self.doc = doc
        
    @staticmethod
    def compare_images(img1, img2, parameter=2000):
        if img1.size > img2.size:
            img2 = img2.resize(img1.size, Image.Resampling.LANCZOS)
        else:
            img1 = img1.resize(img2.size, Image.Resampling.LANCZOS)

        img1 = np.array(img1)
        img2 = np.array(img2)

        mse_score = mse(img1, img2)
        if mse_score < parameter:
            return True
        else:
            return False

    @staticmethod # Print document scan progress
    def print_progress_bar(iteration, total, prefix = '', suffix = '', decimals = 1, length = 100, fill = '█', printEnd = "\r"):
        percent = ("{0:." + str(decimals) + "f}").format(100 * (iteration / float(total)))
        filledLength = int(length * iteration // total)
        bar = fill * filledLength + '-' * (length - filledLength)
        print(f'\r{prefix} |{bar}| {percent}% {suffix}', end = printEnd)
        # Print New Line on Complete
        if iteration == total: 
            print()

    def adjust_pdf(self, rotate=False):
        new_pdf = fitz.Document()
[...]

PR #2303, opened 2023-03-25
Fixing #2300

JorjMcKie:
Page method `bound()` wraps MuPDF function fz_bound_page, but did not catch possible exceptions raised by this function. In these (rare) cases the Python interpreter will end. We now wrap the function with try / catch but do not raise an exception - instead the MuPDF error message is displayed to sys.stderr. For non-PDF documents the infinite rectangle is returned, for PDFs the page rectangle is instead derived from the CropBox as fallback.

### DIFF
--- a/fitz/fitz.i
+++ b/fitz/fitz.i
@@ -1863,29 +1863,6 @@ struct Document
             else Py_RETURN_FALSE;
         }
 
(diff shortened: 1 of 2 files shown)


julian-smith-artifex-com:
Applied manually.

PR #2321, opened 2023-04-04
Fixed search string variable from "word" to "text"

maxhuffy:
The mark_word example print used a variable that is not defined.


maxhuffy:
recheck

Issue #2365, opened 2023-04-21
PyMuPDF 1.22.1 version is producing Type Error while extracting drawings from PDFs


SirishaGorasa:
Hello,

**I am trying to extract drawings using 'page.get_drawings()' and also using 'page.get_cdrawings()' from a PDF.**

I am getting the below error while using page.get_drawings()
![image](https://user-images.githubusercontent.com/47634813/233572420-d0972bba-7a0e-4a98-92e9-37a421371378.png)

And the below error with page.get_cdrawings()
![image](https://user-images.githubusercontent.com/47634813/233572451-c4b5e6b3-cc37-4f36-92cc-9d385712f47b.png)
**Both of them indicate a Type Error. May I know why is this happening ??**

JorjMcKie:
You probably get this error when trying to **_draw_** - not when **_extracting_** drawings.
This is due to a change pertaining to compatibility of joins of a "fill" and a "stroke" path. As opposed to the previous behaviour, value irrelevant to either "fill" or "stroke" are now set to `None`, like `fill_opacity` will be `None` in a stroke path.

SirishaGorasa:
Thanks @JorjMcKie, I fixed it now by adding an existence check. But unfortunately, I now see overlap/bold borders for rectangles which I have created using the output of get_cdrawings(). E.g.: seems like a rectangle drawn again over it.

JorjMcKie:
Found the problem:
All fill-stroke paths are generated automatically by PyMuPDF by joining a **_fill_** followed by a **_stroke_** under certain conditions.
In the most recent version, I forgot to remove the old defalt values `None` for fill color and fill opacity. So fill-strokes had `None` for these two key/values.
This has been fixed and should be published with the next version.

SirishaGorasa:
Thanks for resolving the issue. May I know if this is fixed in the version 1.22.3??

JorjMcKie:
> Thanks for resolving the issue. May I know if this is fixed in the version 1.22.3??

No, not yet, sorry.

SirishaGorasa:
Okay, fine. May I know when can we expect this issue being resolved?

JorjMcKie:
> Okay, fine😊. May I know when can we expect a version with this issue resolved?

The new version is in the making. Currently there is no date fixed.

SirishaGorasa:
Sure. Thanks a lot for quick response.

julian-smith-artifex-com:
[...]

PR #2371, opened 2023-04-25
Update fitz.i related to #2365

JorjMcKie:
This pertains to Page method `get_drawings()`. 
The set of keys for "fs" paths already is complete by definition - so only fill and stroke paths require special attention. Each type should have the complete list of keys as a "fs" path would have, so we generate `None` values where necessary for "f" and "s" paths only. In view of performance, this is probably faster then subclassing `dict` and giving it the `__missing__()` method to achieve the same thing.

Unrelated to this fix:
When dealing with Optional Content Properties, the array of `/Locked` optional content groups is now also supported for input and output.


julian-smith-artifex-com (approved):

fitz/helper-fields.i
```diff
@@ -471,10 +471,10 @@ void JM_get_widget_properties(fz_context *ctx, pdf_annot *annot, PyObject *Widge
         SETATTR_DROP(Widget, "field_type_string",
                 JM_UnicodeFromStr(JM_field_type_text(field_type)));
 
-        #if FZ_VERSION_MAJOR == 1 && FZ_VERSION_MINOR <= 22
+        #if FZ_VERSION_MAJOR == 1 && FZ_VERSION_MINOR < 22
```
julian-smith-artifex-com, line 474: Are you sure this change is correct? We run daily tests with MuPDF 1.21, 1.22 and master (which has version 1.23), and they've been building and passing ok for a while. Maybe your mupdf master needs updating to get the change of its version to 1.23?

fitz/helper-fields.i
```diff
@@ -668,10 +677,10 @@ void JM_set_widget_properties(fz_context *ctx, pdf_annot *annot, PyObject *Widge
     value = GETATTR("field_name");
     if (value != Py_None) {
         char *name = JM_StrAsChar(value);
-        #if FZ_VERSION_MAJOR == 1 && FZ_VERSION_MINOR <= 22
+        #if FZ_VERSION_MAJOR == 1 && FZ_VERSION_MINOR < 22
             char *old_name = pdf_field_name(ctx, annot_obj);
         #else
-            char *old_name = pdf_load_field_name(ctx, annot_obj);
+        char *old_name = pdf_load_field_name(ctx, annot_obj);
```
julian-smith-artifex-com, line 683: Similar comment to above.

julian-smith-artifex-com (commented):
.

fitz/helper-fields.i
JorjMcKie: I only did this because of compile errors.

julian-smith-artifex-com: Great.

Issue #2430, opened 2023-05-30
Incorrectly reducing ref count of Py_None


JorjMcKie:
_**Please provide all mandatory information!**_

## Describe the bug (mandatory)
Starting with version 1.22.0, creating the dictionary `Font.flags` will erroneously decrement the reference count of the "immortal" Python object `Py_None`. After (many) creations, that reference count will eventually reach zero, crashing the interpreter.

## To Reproduce (mandatory)
Create a `Font` object and build `Font.flags` repeatedly.

## Your configuration (mandatory)
 - All operating systems and Python versions
 - PyMuPDF version 1.22.0 and later

JorjMcKie:
Problem fixed and should be rolled out in one of the next versions.

julian-smith-artifex-com:
Fixed in 1.22.5.

PR #2431, opened 2023-05-30
Fixing #2430

JorjMcKie:
Pass a new `None` object to building the `Font.flags` dictionary. We previously directly used `Py_None` which causes a reduction of this "immortal" Python object's ref count.


JorjMcKie:
This should go into the next version if at all possible.
A test script should not be required, as the change is obvious really.

julian-smith-artifex-com (approved):

JorjMcKie (commented):
just a space before the None paramezter

PR #2528, opened 2023-07-11
Extend sRGB functions

JorjMcKie:
Add a new function that converts an sRGB integer to its string equivalent as used by html, e.g. yellow, `0xFFFF00` returns `"#FFFF00"`.


JorjMcKie:
The code change comprises a new "standalone" function. No existing code / behaviour changed in any way. So the failure must be due to the point in time **_when_** the tests were run.

docs/functions.rst
```diff
@@ -567,13 +582,13 @@ Yet others are handy, general-purpose utilities.
...
-      4. For data compactness, the character's unicode is provided here. Use built-in function `chr()` for the character itself.
+      3. Line width in this context is important only for processing `span["type"] != 0` i.e. characters are stroked and not filled: it determines the thickness of the character's border line. This value may not be provided at all with the text data. In this case, a value of 5% of the fontsize (`span["size"] * 0,05`) is generated. Often, an "artificial" bold text in PDF is created by `2 Tr`. There is no equivalent span type for this case. Instead, respective text is represented by two consecutive spans -- which are identical in every aspect, except for their types, which are 0, resp 1. It is your responsibility to handle this type of situation - in :meth:`Page.get_text`, MuPDF is doing this for you.
```
jamie-lemon, line 585: while we are here can we update resp 1 to : respectively 1 ?

jamie-lemon (changes requested):

docs/functions.rst
jamie-lemon: All docs update LGTM aside from this one tiny tweak.

JorjMcKie: Done: "resp" -> "respectively"

jamie-lemon (approved):

PR #2610, opened 2023-08-22
Fixing typos in documentation and missing parameters description

ellacroix:
I fixed two small typos in Page documentation ("flags (in)" -> "flags (int)").
While browsing the package code i found that some functions to extract text from a TextPage allowed the use of a sort parameter that wasn't specified in the documentation, so i added it to the functions definitions and their description. I used the description on the sort parameter description from Page.get_text().

I also found an error that i can't fix via pull request: The [documentation for TextPage](https://pymupdf.readthedocs.io/en/latest/textpage.html) specify that we can use TextPage.extractText() or  TextPage.extractTEXT(), but only one of those functions exist in the package (TextPage.extractText()).
The easiest fix would be to add TextPage.extractTEXT() in the code on your side, as a copy of the existing function.
If not, i could just erase TextPage.extractTEXT() from the documentation in this pull request.

### DIFF
--- a/docs/page.rst
+++ b/docs/page.rst
@@ -1086,7 +1086,7 @@ In a nutshell, this is what you can do with PyMuPDF:
 
       Create a :ref:`TextPage` for the page.
 
-      :arg in flags: indicator bits controlling the content available for subsequent text extractions and searches -- see the parameter of :meth:`Page.get_text`.
+      :arg int flags: indicator bits controlling the content available for subsequent text extractions and searches -- see the parameter of :meth:`Page.get_text`.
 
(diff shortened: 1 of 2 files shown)


ellacroix:
Another fix i thought of would be to add this code to \_\_init\_\_.py, the downside being that auto-completion tool like Intellisense won't propose the function for fitz.TextWriter

```
# ------------------------------------------------------------------------
# TextPage
# ------------------------------------------------------------------------
fitz.TextPage.extractTEXT = fitz.TextPage.extractText
```

jamie-lemon:
Seems like we use `extractTEXT` as a synonym for `extractText` - so I think this is fine.

JorjMcKie (approved):

Issue #2593, opened 2023-08-14
get_pixmap with alpha produce gray image for transparent white image


mamome2021:
_**Please provide all mandatory information!**_

## Describe the bug (mandatory)
I have a 50% transparent white image in pdf file. I want to use `Page.get_pixmap(alpha=True)` to render the page as pixmap, convert pixmap to pillow image, and paste it on another image (as watermark, for example). But the pixmap I get is gray image.

## To Reproduce (mandatory)

code snippet:
```
import fitz
from PIL import Image

doc=fitz.open('alpha.pdf')
page=doc[0]
pix=page.get_pixmap(alpha=True)
img=Image.frombytes('RGBA', [pix.width, pix.height], pix.samples)
img.show()
img.save('alpha.tiff')
```

example pdf:
[alpha.pdf](https://github.com/pymupdf/PyMuPDF/files/12334431/alpha.pdf)

## Expected behavior (optional)
white image with transparency

## Your configuration (mandatory)
 - Operating system, potentially version and bitness
 - Python version, bitness
 - PyMuPDF version, installation method (**wheel** or **generated** from source).

3.11.3 (main, Jun  5 2023, 09:32:32) [GCC 13.1.1 20230429] 
 linux 
 
PyMuPDF 1.22.5: Python bindings for the MuPDF 1.22.2 library.
Version date: 2023-06-21 00:00:01.
Built for Python 3.11 on linux (64-bit).
PyMuPDF generated from source

## Additional context (optional)
If I save pixmap with pix.save(), the saved image looks fine

JorjMcKie:
This effect is caused by setting transparency to 50% for apparently all your drawings.

mamome2021:
Do you have suggestion for keeping the image white? I have tried 
```
img=Image.open(BytesIO(pix.tobytes()))
```
as alternative, but this is 50 times slower.

JorjMcKie:
> Do you have suggestion for keeping the image white? I have tried
> 
> ```
> img=Image.open(BytesIO(pix.tobytes()))
> ```
> 
> as alternative, but this is 50 times slower.

It also doesn't solve the problem or does it?
No, the error happened when the PDF was **_created._** Each of the vector graphics objects refers to an `/ExtGState` object that looks like this:
```
<</CA 0.5
   /ca 0.5
>>
```
[...]

PR #2632, opened 2023-08-29
Support Pillow mode "RGBa"

JorjMcKie:
For Pixmap image output via Pillow, also support mode RGBa with the new keyword parameter "unmultiply". The Pillow image will then be created with unmultiplied pixels.

This fix also addresses the closed issue #2593: Pixmaps as a result of page rendering can now be converted to Pillow images with unmultiplied alpha values.


JorjMcKie:
Note: no dedicated test script necessary here.

julian-smith-artifex-com (approved):

PR #2633, opened 2023-08-29
Support pixmap "unmultiply"

JorjMcKie:
For Pillow output of RGB images with transparency, the new parameter `unmultiply=True` causes the use of Pillow mode "RGBa" (instead of "RGBA"). The Pillow image will then be created with unmultiplied pixels.

This is the documentary aspect of support closed issue #2593.


jamie-lemon:
LGTM

Issue #2556, opened 2023-07-25
Segmentation fault at caling get_cdrawings(extended=True)


pulsar314:
## Description
If a document contains sequences like
```
q
100 100 m
W n
```
invocation of `page.get_cdrawings(extended=True)` results in a segfault.

In this case `dev_pathdict` is being cleared due to the empty commands list, but `jm_lineart_clip_path` and `jm_lineart_clip_stroke_path` are trying to get a value from the dict.

## Configuration
 - Linux MINT
 - Python 3.8
 - PyMuPDF 1.22.5, installed via pip

pulsar314:
Unfortunately, I cannot provide you with the original PDF, but there is a reproduction of the failing sequence
[segfault.pdf](https://github.com/pymupdf/PyMuPDF/files/12182681/segfault.pdf)

julian-smith-artifex-com:
Fixed in 1.23.4.

Issue #2553, opened 2023-07-20
Invalid characters in versions >= 1.22


brandenkmurray:
## Describe the bug (mandatory)
`get_text()` in versions >=1.22 produces � characters in some cases, usually related to LaTex. This was not an issue in `v1.21.1` and other PDF libraries extract the text just fine (though `pdfplumber` appears to miss a few characters)

Additionally, `get_text(sort=True)` converts the � to `\udc52` which creates other issues e.g. causes `print()` to fail with error `UnicodeEncodeError: 'utf-8' codec can't encode character '\udc52' in position 35: surrogates not allowed`

[1001.2481.pdf](https://github.com/pymupdf/PyMuPDF/files/12113342/1001.2481.pdf)

## To Reproduce (mandatory)
```
import fitz
import pdftotext
import pdfplumber

def print_comparison(fn, page):
    #pymupdf
    pymupdf_doc = fitz.open(fn)

    #pdftotext
    with open(fn, "rb") as f:
        pdftotext_doc = pdftotext.PDF(f)

    #pdfplumber
    pdfplumber_doc = pdfplumber.open(fn)

    print("PyMuPDF:\n")
    print(repr(pymupdf_doc[page].get_text()))
    print("\npdftotext:\n")
    print(repr(pdftotext_doc[page]))
    print("\npdfplumber:\n")
    print(repr(pdfplumber_doc.pages[page].extract_text()))


print_comparison('1001.2481.pdf', 10)
```

```
PyMuPDF:

' \n \nFig. 3. Growth rate of a turbulent spot close to the phase transition. Square of the \ndimensionless growth rate of turbulent spots plotted as a function of Re close to the \nphase transition in pipe (A), channel (B) and square duct (C). The line follows \n𝐺 ∝  𝑅𝑐��� − 𝑅𝑐���𝐶 1/2 which yields the following critical numbers: 𝑅𝑐���𝐶\n𝑝𝑖𝑝𝑐��� = 2550, \n 𝑅𝑐���𝐶\n𝑐ℎ𝑎𝑛𝑛𝑐���𝑙 = 1480  and  𝑅𝑐���𝐶\n𝑐���𝑡���𝑐𝑡 = 2250 . Note that for the channel non-linear \nsaturation sets in much earlier than in the other two cases.  \n \n \n \n'

pdftotext:

[...]

Issue #2522, opened 2023-07-10
Typo in set_layer() - NameError: name 'f' is not defined


claudinecombe:
_**Please provide all mandatory information!**_

## Describe the bug (mandatory)

I am trying to use `rbgroups` in the `set_layer()` function but it throws an error
```
  File "fitz/fitz.py", line 5189, in set_layer
    if f != set():
NameError: name 'f' is not defined
```

When looking at the code, `f` should be replaced by `s`

```
        if rbgroups:
            if type(rbgroups) not in (list, tuple):
                raise ValueError("bad type: 'rbgroups'")
            for x in rbgroups:
                if not type(x) in (list, tuple):
                    raise ValueError("bad RBGroup '%s'" % x)
                s = set(x).difference(ocgs)
                if f != set():
                    raise ValueError("bad OCGs in RBGroup: %s" % s)
```

## To Reproduce (mandatory)
Explain the steps to reproduce the behavior, For example, include a minimal code snippet, example files, etc.

```
        doc = fitz.open()
        xrefs = {}
        xrefs["oc1"] = doc.add_ocg("oc1", on=False)
        xrefs["oc2"] = doc.add_ocg("oc2", on=True)
        xrefs["oc3"] = doc.add_ocg("oc3", on=False)
        xrefs["oc4"] = doc.add_ocg("oc4", on=False)
        doc.set_layer(-1, on=[3], off=[4, 5, 6], rbgroups=[[4, 5, 6]])
```

JorjMcKie:
Thank you!
There indeed is a typo in that line. Will be fixed in next version.
That line should read `if s != set():`

julian-smith-artifex-com:
Fixed in 1.23.5.

PR #2661, opened 2023-09-11
Word delimiter support, fixes #2637,  #2556, #2553, #2522

### LINKED ISSUE #2637: Page.insert_textbox incorrectly handles the last word if it starts a new line
_**Please provide all mandatory information!**_

## Describe the bug (mandatory)
It seems that insert_textbox doesn't handle different fonts properly. Even when i add a default font, the text fits not always perfectly within the textbox. However, with a different font, there's either too much whitespace or the text overflows beyond the textbox.

The while loop will stop because the rc value is usually greater than 0, ( it is the "left white space value" right?) often around 2. However, with some fonts (like Allura), the textbox has almost half of its space left as whitespace.

So the rc value should be much higher.

## To Reproduce (mandatory)
Explain the steps to reproduce the behavior, For example, include a minimal code snippet, example files, etc.

 ```
newdoc[r_frame["page"]-1].insert_font(fontname="Text_font", fontfile="Files/Fonts/" + order["font"])
                
                font="Text_font"
                text_color = (0/255, 38/255, 145/255)

             
                size=order["min-font-size"]
              
                rc= -1
                
         
                while rc < 0 :#or (0 < rc < 10):
                   
                    
                    #print(size)
                    size=size-1
                    rc = newdoc[r_frame["page"]-1].insert_textbox(rect, text,
                        fontsize=size,
                        fontname=font,
                        rotate = 0,
                        color=text_color)
                    print(rc)
```

## Expected behavior (optional)
Describe what you expected to happen (if not obvious).
The text should fit neatly within the rectangle, especially at the bottom. Some whitespace is expected due to varying fonts, but the text shouldn't extend outside the rectangle.

## Screenshots (optional)
See screenshot, for outcome off different fonts.

Example different fonts

JorjMcKie:
For text extraction `get_text("words")`, or extractWORDS, words are defined as strings not containing white space. This change allows adding up to 64 characters to also function as delimiters. This allows for instance to separate words from punctuations or to decompose an e-mail address into its components.

Other changes:

Fixing #2522: correcting the typo

Remove some unnecessary setting of flags when creating annotations.

Fixing #2553:
Adjust plain text extraction to use the same approach as other variants. This entails using Unicode escape strings on output instead of using the output of fz_chartorune. Another consequence is that standard text output is directed to a fz_buffer instead to a fz_output.

Fixing #2556: Add checking the existence of path dictionaries at every possible place. Includes an additional test function.

Add functions JM_ignore_rect / JM_ignore_irect which return a bool. The functions return True if the rectangle should be ignored. This is the case for infinite and empty rectangles, but also for any rectangle that has a common edge with the infinite rectangle.

Support variable setting of character border widths for insert_text() / insert_textbox(). This is a factor to be multiplied with the font size. Default is 0.05 (read: 5% of the fontsize). This value is relevant for text rendering modes 1 and 2 only.

Fixing #2637:
In Page.insert_textbox, when the last word of a line won't fit in the line buffer, we did not increase the line position. This is now handled correctly.

### DIFF
--- a/fitz/fitz.i
+++ b/fitz/fitz.i
@@ -4030,7 +4030,7 @@ if rbgroups:
         if not type(x) in (list, tuple):
             raise ValueError("bad RBGroup '%s'" % x)
         s = set(x).difference(ocgs)
-        if f != set():
+        if s != set():
             raise ValueError("bad OCGs in RBGroup: %s" % s)
 
 if basestate:
@@ -5542,8 +5542,6 @@ struct Page {
                 r = pdf_annot_rect(gctx, annot);
                 r = fz_make_rect(p.x, p.y, p.x + r.x1 - r.x0, p.y + r.y1 - r.y0);
                 pdf_set_annot_rect(gctx, annot, r);
-                int flags = PDF_ANNOT_IS_PRINT;
-                pdf_set_annot_flags(gctx, annot, flags);
 
                 if (icon)
                     pdf_set_annot_icon_name(gctx, annot, icon);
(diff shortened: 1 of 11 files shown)


JorjMcKie:
@julian-smith-artifex-com - I also made the corresponding "rebased" fixes. I hope I spotted them all ...

PR #2679, opened 2023-09-21
Adds a fontsize definition to the Glossary and updates relevant links.

jamie-lemon:


JorjMcKie (commented):
I found an error (my error) in annot.rst:
Line 239: I am talking about "190 pixels" - which is wrong. Should be "190 points".

Apart from that: OK!

JorjMcKie (approved):

PR #2688, opened 2023-09-25
Clarify use of encoding parameter

JorjMcKie:
Improve description of "encoding" parameter in "insert_text()". Clarify that using file-based fonts is the better alternative to support languages like Russian or Greek.


jamie-lemon:
LGTM

PR #2698, opened 2023-09-28
Flexible word delimiters

JorjMcKie:
This change offers to define up to 64 additional characters for use as delimiters in page.get_text("words"). Delimiters can be set either temporarily in the method itself, or permanently for all subsequent word extractions.

Please note that in rebased, no respective changes were made in extra.i. Any speed boosting will need extra code there.

Other changes:

* plain text output characters are now handled the same way as in all other extraction variants. This entails using fz_buffer as intermediate storage for extracted text - not any longer fz_output.

* in rare cases, fz_stext_page returns invalid bboxes for images or text: not empty, not infinite, but still with one or more of its edges having coordinates in common with the infinite rectangle. We define a new function "JM_ignore_rect()" to allow filtering out (ignoring) them.

* finally fixed typo in issue #2522

* added TOOLS.store_shrink() to Document method "reload_page()" to ensure the absence of cached, now possibly invalid objects pertaining to this page.

* When adding annotation to a page, removed unnecessary (possibly even wrong) extra annotation flag settings, so now MuPDF's defaults will prevail.

* Added flexibility to border_width parameter in Page methods "insert_text"/"insert_textbox". This is now interpreted as a fraction of font size, such that e.g. a value of 0.05 corresponds to a border width of 5% of the font size.

### DIFF
--- a/fitz/fitz.i
+++ b/fitz/fitz.i
@@ -4030,7 +4030,7 @@ if rbgroups:
         if not type(x) in (list, tuple):
             raise ValueError("bad RBGroup '%s'" % x)
         s = set(x).difference(ocgs)
-        if f != set():
+        if s != set():
             raise ValueError("bad OCGs in RBGroup: %s" % s)
 
 if basestate:
@@ -4625,6 +4625,7 @@ if basestate:
                     old_annots[k] = v
                 page._erase()  # remove the page
                 page = None
+                TOOLS.store_shrink(100)
                 page = self.load_page(pno)  # reload the page
 
                 # copy annot refs over to the new dictionary
@@ -5542,8 +5543,6 @@ struct Page {
                 r = pdf_annot_rect(gctx, annot);
                 r = fz_make_rect(p.x, p.y, p.x + r.x1 - r.x0, p.y + r.y1 - r.y0);
                 pdf_set_annot_rect(gctx, annot, r);
-                int flags = PDF_ANNOT_IS_PRINT;
-                pdf_set_annot_flags(gctx, annot, flags);
 
                 if (icon)
                     pdf_set_annot_icon_name(gctx, annot, icon);
@@ -5554,7 +5553,6 @@ struct Page {
                 pdf_dict_put_text_string(gctx, annot_obj, PDF_NAME(Contents), filename);
                 pdf_update_annot(gctx, annot);
                 pdf_set_annot_rect(gctx, annot, r);
-                pdf_set_annot_flags(gctx, annot, flags);
                 JM_add_annot_id(gctx, annot, "A");
             }
             fz_always(gctx) {
@@ -7616,10 +7614,9 @@ def insert_font(self, fontname="helv", fontfile=None, fontbuffer=None,
                 annot_xrefs = [a[0] for a in self.annot_xrefs() if a[1] not in skip_types]
             else:
                 annot_xrefs = [a[0] for a in self.annot_xrefs() if a[1] in types and a[1] not in skip_types]
+
             for xref in annot_xrefs:
-                annot = self.load_annot(xref)
-                annot._yielded=True
-                yield annot
+                yield self.load_annot(xref)
 
 
         def widgets(self, types=None):
@@ -11533,13 +11530,18 @@ struct TextPage {
         extractIMGINFO(int hashes=0)
         {
             fz_stext_block *block;
-            int block_n = -1;
+            int block_n = -1, leave = 0;
             fz_stext_page *this_tpage = (fz_stext_page *) $self;
             PyObject *rc = NULL, *block_dict = NULL;
             fz_pixmap *pix = NULL;
+            fz_rect bbox;
             fz_try(gctx) {
                 rc = PyList_New(0);
                 for (block = this_tpage->first_block; block; block = block->next) {
+                    bbox = block->bbox;
+                    if (JM_ignore_rect(bbox)) {
+                        continue; // guard against nonsense block bbox
+                    }
                     block_n++;
                     if (block->type == FZ_STEXT_BLOCK_TEXT) {
                         continue;
@@ -11556,7 +11558,7 @@ struct TextPage {
                     block_dict = PyDict_New();
                     DICT_SETITEM_DROP(block_dict, dictkey_number, Py_BuildValue("i", block_n));
                     DICT_SETITEM_DROP(block_dict, dictkey_bbox,
-                                    JM_py_from_rect(block->bbox));
+                                    JM_py_from_rect(bbox));
                     DICT_SETITEM_DROP(block_dict, dictkey_matrix,
                                     JM_py_from_matrix(block->u.i.transform));
                     DICT_SETITEM_DROP(block_dict, dictkey_width,
@@ -11695,7 +11697,7 @@ struct TextPage {
             fz_rect wbbox = fz_empty_rect;  // word bbox
             fz_stext_page *this_tpage = (fz_stext_page *) $self;
             fz_rect tp_rect = this_tpage->mediabox;
-
+            int word_delimiter = 0;
             PyObject *lines = NULL;
             fz_try(gctx) {
                 buff = fz_new_buffer(gctx, 64);
@@ -11717,10 +11719,12 @@ struct TextPage {
                                 !fz_is_infinite_rect(tp_rect)) {
                                 continue;
                             }
-                            if (ch->c == 32 && buflen == 0)
-                                continue;  // skip spaces at line start
-                            if (ch->c == 32) {
-                                if (!fz_is_empty_rect(wbbox)) {
+                            word_delimiter = JM_is_word_delimiter(ch->c);
+                            if (word_delimiter) {  // encountered end of word
+                                if (buflen == 0) {
+                                    continue;
+                                }
+                                if (!fz_is_empty_rect(wbbox)) {  // output word
                                     word_n = JM_append_word(gctx, lines, buff, &wbbox,
                                                         block_n, line_n, word_n);
                                 }
@@ -11806,10 +11810,10 @@ struct TextPage {
                         fz_print_stext_page_as_xhtml(gctx, out, this_tpage, 0);
                         break;
                     default:
-                        JM_print_stext_page_as_text(gctx, out, this_tpage);
+                        JM_print_stext_page_as_text(gctx, res, this_tpage);
                         break;
                 }
-                text = JM_UnicodeFromBuffer(gctx, res);
+                text = JM_EscapeStrFromBuffer(gctx, res);
 
             }
             fz_always(gctx) {
@@ -12168,6 +12172,7 @@ struct TextWriter
             morph: tuple(Point, Matrix), apply a matrix with a fixpoint.
             matrix: Matrix to be used instead of 'morph' argument.
             render_mode: (int) PDF render mode operator 'Tr'.
+            border_width: (float) stroke line Width. Relevant for render mode > 0.
         """
 
         CheckParent(page)
@@ -12185,6 +12190,8 @@ struct TextWriter
             opacity = self.opacity
         if color is None:
             color = self.color
+        if render_mode < 0:
+            render_mode = 0
         %}
 
         %pythonappend write_text%{
@@ -12234,7 +12241,7 @@ struct TextWriter
                 temp = line.split()
                 fsize = float(temp[1])
                 if render_mode != 0:
-                    w = fsize * 0.05
+                    w = fsize * border_width
                 else:
                     w = 1
                 new_cont_lines.append("%g w" % w)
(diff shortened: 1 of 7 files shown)


fitz/fitz.i
```diff
@@ -12185,6 +12190,8 @@ struct TextWriter
             opacity = self.opacity
         if color is None:
             color = self.color
+        if render_mode < 0:
+            render_mode = 0
```
julian-smith-artifex-com, line 12194: Could this be `assert render_mode >= 0` ?

fitz/fitz.i
```diff
@@ -14412,6 +14419,63 @@ struct Tools
...
+        if delims == None:
+            delims = []
+        if not hasattr(delims, "__getitem__") or len(delims) > 64:
+            raise ValueError("bad delimiter value(s)")
+
+        try:
+            delims = set([ord(c) for c in delims if ord(c) > 32])
+        except:
+            print("bad delimiter value(s)")
+            raise
```
julian-smith-artifex-com, line 14434: This ignores delimitors <= 32. Could we instead do `assert c > 32`?
[Or perhaps simply allow all delimiters?]

fitz/fitz.i
```diff
@@ -14412,6 +14419,63 @@ struct Tools
         }
 
 
+        FITZEXCEPTION(set_word_delimiters, !result)
+        %pythonprepend set_word_delimiters %{
+        """Set characters to be word delimiters."""
+        if delims == None:
+            delims = []
+        if not hasattr(delims, "__getitem__") or len(delims) > 64:
+            raise ValueError("bad delimiter value(s)")
```
julian-smith-artifex-com, line 14428: Seems a bit of a shame to have a fixed-size buffer for this.

fitz/fitz.i
```diff
@@ -14412,6 +14419,63 @@ struct Tools
...
+
+        try:
+            delims = set([ord(c) for c in delims if ord(c) > 32])
+        except:
```
julian-smith-artifex-com, line 14432: Minor thing, but doing `except Exception:` ignores Ctrl-C etc, allowing user to interrupt.

fitz/fitz.i
```diff
@@ -14412,6 +14419,63 @@ struct Tools
...
+            int i, len = (int) PyTuple_Size(delims);
+            if (!len) {
+                word_delimiters[0] = 0; // set list to empty
+                return Py_False;
+            }
+
+            fz_try(gctx) {
+                for (i = 0; i < len; i++) {
+                    word_delimiters[i] = (int) PyLong_AsLong(PyTuple_GET_ITEM(delims, (Py_ssize_t) i));
+                    word_delimiters[i+1] = 0;
```
julian-smith-artifex-com, line 14448: Could we put `word_delimiters[len] = 0` outside the loop?

fitz/helper-geo-c.i
```diff
@@ -76,8 +76,8 @@ JM_rect_from_py(PyObject *r)
 
     for (i = 0; i < 4; i++) {
         if (JM_FLOAT_ITEM(r, i, &f[i]) == 1) return fz_infinite_rect;
-        if (f[i] < FZ_MIN_INF_RECT) f[i] = FZ_MIN_INF_RECT;
-        if (f[i] > FZ_MAX_INF_RECT) f[i] = FZ_MAX_INF_RECT;
+        if (f[i] <= FZ_MIN_INF_RECT) f[i] = FZ_MIN_INF_RECT;
+        if (f[i] >= FZ_MAX_INF_RECT) f[i] = FZ_MAX_INF_RECT;
```
julian-smith-artifex-com, line 80: I think these changes have no effect...?

fitz/helper-geo-c.i
```diff
@@ -105,8 +131,8 @@ JM_irect_from_py(PyObject *r)
 
     for (i = 0; i < 4; i++) {
         if (JM_INT_ITEM(r, i, &x[i]) == 1) return fz_infinite_irect;
-        if (x[i] < FZ_MIN_INF_RECT) x[i] = FZ_MIN_INF_RECT;
-        if (x[i] > FZ_MAX_INF_RECT) x[i] = FZ_MAX_INF_RECT;
+        if (x[i] <= FZ_MIN_INF_RECT) x[i] = FZ_MIN_INF_RECT;
+        if (x[i] >= FZ_MAX_INF_RECT) x[i] = FZ_MAX_INF_RECT;
```
julian-smith-artifex-com, line 135: This change has no effect?

fitz/helper-stext.i
```diff
@@ -73,11 +92,14 @@ JM_new_stext_page_ocr_from_page(fz_context *ctx, fz_page *page, fz_rect rect, in
...
 //---------------------------------------------------------------------------
 // APPEND non-ascii runes in unicode escape format to fz_buffer
+// NOTE: As a consequence, fz_chartorune is NOT USED.
 //---------------------------------------------------------------------------
 void JM_append_rune(fz_context *ctx, fz_buffer *buff, int ch)
 {
     if ((ch >= 32 && ch <= 255) || ch == 10) {
         fz_append_byte(ctx, buff, ch);
+    } else if (ch >= 0xd800 && ch <= 0xdfff) {
+        fz_append_string(ctx, buff, "\\ufffd");
```
julian-smith-artifex-com, line 102: Could do with a comment explaining what these values are.

fitz/utils.py
```diff
@@ -536,25 +536,36 @@ def get_text_words(
     flags: OptInt = None,
     textpage: TextPage = None,
     sort: bool = False,
+    delimiters=None,
 ) -> list:
     """Return the text words as a list with the bbox for each word.
 
     Args:
         flags: (int) control the amount of data parsed into the textpage.
+        delimiters: (str,list) characters to use as word delimiters
```
julian-smith-artifex-com, line 545: delimiters are _extra_ characters - i think we always treat 32 and 160 as delimiters?

fitz/utils.py
```diff
@@ -536,25 +536,36 @@ def get_text_words(
...
         raise ValueError("not a textpage of this page")
+    if delimiters is not None:
+        TOOLS.set_word_delimiters(delimiters)
     words = tp.extractWORDS()
     if textpage is None:
         del tp
     if sort is True:
         words.sort(key=lambda w: (w[3], w[0]))
+    if delimiters is not None:
+        TOOLS.set_word_delimiters(old_delimiters)
```
julian-smith-artifex-com, line 568: Not thread-safe. Ok for classic, but we might want to avoid this sort if thing in rebased.

fitz/utils.py
```diff
@@ -1725,7 +1742,7 @@ def insert_textbox(
     align: int = 0,
     rotate: int = 0,
     render_mode: int = 0,
-    border_width: float = 1,
+    border_width: float = 0.05,
```
julian-smith-artifex-com, line 1745: Why this change?

src/__init__.py
```diff
@@ -96,6 +96,8 @@ def get_env_bool( name, default):
 # Unset ascender / descender corrections
 g_skip_quad_corrections = 0
 
+# additional word delmiters
+g_word_delimiters = [0] * 65
```
julian-smith-artifex-com, line 100: Feels like we should do `g_word_delimiters = list()`. Or, better, pass around as a fn param so things are thread safe.

src/__init__.py
```diff
@@ -5266,7 +5269,7 @@ def set_layer(self, config, basestate=None, on=None, off=None, rbgroups=None, lo
                 if not type(x) in (list, tuple):
                     raise ValueError("bad RBGroup '%s'" % x)
                 s = set(x).difference(ocgs)
-                if f != set():
+                if s != set():
```
julian-smith-artifex-com, line 5272: Is this a fix for a particular issue?

src/__init__.py
```diff
@@ -13906,12 +13914,28 @@ def JM_append_rune(buff, ch):
     '''
     if (ch >= 32 and ch <= 255) or ch == 10:
         mupdf.fz_append_byte(buff, ch)
+    elif ch >= 0xd800 and ch <= 0xdfff:
+        mupdf.fz_append_string(buff, "\\ufffd")
```
julian-smith-artifex-com, line 13918: As with classic, would like a comment saying what these character ranges are.

src/__init__.py
```diff
@@ -20992,6 +21040,36 @@ def set_graphics_min_line_width(min_line_width):
         '''
         mupdf.fz_set_graphics_min_line_width(min_line_width)
 
+
+    @staticmethod
+    def get_word_delimiters():
+        """Return extra word delimiting characters."""
+        global g_word_delimiters
+        delims = [chr(c) for c in g_word_delimiters if c != 0]
+        return delims
```
julian-smith-artifex-com, line 21049: Is this right if g_word_delimiters is set to 10 characters, then reset to 5 characters? In this case i think it could be ....0.....0, where `.` is any non-zero character. I.e. we probably want to terminate at the first 0.

Also, don't need `global g_word_delimiters` - only required if we are modifying it.

src/__init__.py
```diff
@@ -20992,6 +21040,36 @@ def set_graphics_min_line_width(min_line_width):
...
+            delims = set([ord(c) for c in delims if ord(c) > 32])
+        except:
+            print("bad delimiter value(s)")
+            raise
+        delims = tuple(delims)
+        if not delims:
+            g_word_delimiters = [0] * 65
+        else:
+            g_word_delimiters = delims
+        return # extra.set_word_delimiters(delims)
```
julian-smith-artifex-com, line 21070: If we can set `g_word_delimiters` to a set here, we probably don't need it to be `[0]*65` when we initiallise?

fitz/fitz.i
JorjMcKie: I dont wan't to allow delimiters that already are ones by their very nature. At the very least, 0x00 cannot be allowed, because it serves as the (premature) end of add'l delimiter array.
Also I want to allow strings and sequences of single characters - therefore I need the try-except.

fitz/fitz.i
JorjMcKie: The idea is to allow 'string.punctuation` as delimiters (32 items) plus any equivalent of those crude Chinese characters like "。" (Unicode 0x3002). Nobody will ever be able to keep track of more than this number. We must not forget that none of the delimiters will ever be part of any extracted word ...

fitz/helper-geo-c.i
JorjMcKie: In Python, larger/smaller integers than that are possible.

fitz/utils.py
JorjMcKie: Yes, read as *extra* delimiters

fitz/utils.py
JorjMcKie: The idea is to allow temporary (extra) delimiters - this call only.

fitz/utils.py
```diff
@@ -3755,99 +3773,107 @@ def pixlen(x):
...
-                pos += lheight * progr  # increase line position
 
-        more = (pos - maxpos) * progr  # difference to rect size limit
+        # compute used part of the textbox
+        if text.endswith("\n"):
+            text=text[:-1]
+        lb_count = text.count("\n") + 1 # number of lines written
+
+        # text height = line count * line height plus one descender value
+        text_height = lheight * lb_count - descender * fontsize
```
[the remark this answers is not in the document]
JorjMcKie: descender is a negative value - always.

fitz/fitz.i
julian-smith-artifex-com: In rebased, i think it would actually be simpler to implement and document if we used an unlimited Python list.

fitz/helper-geo-c.i
julian-smith-artifex-com: Yes, but in general `if x <= y: x = y` behaves identically to `if x < y: x = y`, so i don't think this diff has any effect.

JorjMcKie:
Thanks for review! I will give them a closer look today or over the weekend. PBased on it, I probably have to redesign things.

JorjMcKie:
BTW did you understand the reason for the failed test?

julian-smith-artifex-com:
> BTW did you understand the reason for the failed test?

* If you click on `details` on the right of the `Test quick` failure, this will bring up the page for the failing workflow.
* Click on the 6-pointed wheel icon on the top right.
* In the resulting menu, choose `View raw logs`.
* This shows the output of the workflow (showing the PyMuPDF build and running of tests).

In this case, the problem is an error when compiling the swig-generated file fitz/fitz.i.c (this is classic implementation):
```
/project/fitz/fitz.i.c: In function ‘JM_is_word_delimiter’:
/project/fitz/fitz.i.c:8205:10: error: ‘i’ undeclared (first use in this function)
 8205 |     for (i = 0; i < (int) nelem(word_delimiters); i++) {
```
So this is a bug in `fitz/fitz.i`. The line number will be different (unfortunately swig doesn't emit `#line` markers), so you'll have to search for the `for (i = 0; i < (int) nelem(word_delimiters); i++)`.

PR #2711, opened 2023-10-03
Fix #2596

### LINKED ISSUE #2596: save(garbage=3) breaks get_pixmap() with side effect
# Problem
After saving the fitz.Document with `pdf.save("filename.pdf", garbage=3)` the embedded images of the document I can see in the pixmap of the pages are mixed up: `pixmap = pdf[0].get_pixmap()`.
`fitz.Document.save(garbage=3)` and `fitz.Document.write(garbage=3)` appear to have side effects that change the document in place.
When I set `garbage` to 2 it works as expected.

# Observation
Checked checkboxes become unchecked boxes and the other way around, some checkboxes become logos etc..

# Code Example

``` python
pdf = fitz.Document("filename.pdf")
# pixmap of the first page looks just fine
pix0 = pdf[0].get_pixmap()

# save with "garbage=3" breaks the pdf in memory (the saved pdf on disk is still fine)
pdf.save("debug.pdf", garbage=3)

# the pixmap is now broken - images in the pixmap are mixed up
pix1 = pdf[0].get_pixmap()  # <---------- PROBLEM

# read the saved pdf again
new_pdf = fitz.Document("debug.pdf")
# the pixmap of the saved and read pdf file is not broken!
pix2 = new_pdf[0].get_pixmap()
```

Sadly I can't provide the affected  internal documents.

# Version
```
PyMuPDF 1.20.2: Python bindings for the MuPDF 1.20.3 library.
Version date: 2022-08-13 00:00:01.
Built for Python 3.9 on darwin (64-bit).
```
Installed with `pip install PyMuPdf`
I also tested it on 1.22.5 on macOS with the same result.

JorjMcKie:
When continuing to use a page object after the PDF has been saved with garbage options may change the MuPDF cache such that e.g. rendering the page after the save will produce wrong appearance or even garbled appearance. The user needs to reload the page (page = doc.reload(page)) before using it further, especially for rendering. The reload method now always empties the cache.


julian-smith-artifex-com (approved):

JorjMcKie:
Thanks!

Issue #2548, opened 2023-07-17
Fitz freezes on some PDFs when calling the fitz.Page.get_text_blocks method.


AndreyRomanyukov:
## Describe the bug (mandatory)
Fitz freezes on some PDFs when calling the fitz.Page.get_text_blocks method.

## To Reproduce (mandatory)
Download the pdf that causes a freeze.

original https://aacr.figshare.com/articles/journal_contribution/Supplementary_Data_from_Targeting_Therapeutic_Resistance_and_Multinucleate_Giant_Cells_in_CCNE1-Amplified_HR-Proficient_Ovarian_Cancer/22523824/1/files/39986620.pdf
mirror https://www.dropbox.com/s/s7zjp7a8ys5ibh0/mct-21-0873_supplementary_data_s1_supps1.pdf?dl=0

Run the python code

```
from io import BytesIO

import fitz


path = 'mct-21-0873_supplementary_data_s1_supps1.pdf'

with open(path, 'rb') as opened:
    stream = BytesIO(opened.read())

with fitz.open(stream=stream, filetype="pdf") as pdf:
    for page in pdf:
        print(page)
        blocks = page.get_text_blocks()
        print('blocks:', len(blocks))
```

The program will print

```
page 0 of <memory, doc# 1>
blocks: 2
page 1 of <memory, doc# 1>
blocks: 2
page 2 of <memory, doc# 1>
```

and then freeze.

## Additional context (optional)
Reproduces on PyMuPDF==1.22.3 and PyMuPDF==1.22.5. Reproduces on macOS 12.6.5 and Ubuntu 20.04.2

JorjMcKie:
The problem occurs when extracting text in general. It more precisely is loop.
It is also not a problem inside PyMuPDF but in the base library MuPDF.
`mutool draw -o test.txt file.pdf` also loops.
I will forward the problem to MuPDF's issue system.

AndreyRomanyukov:
> I will forward the problem to MuPDF's issue system.

Thanks!

JorjMcKie:
Sorry for the delay, here is the MuPDF bug report: https://bugs.ghostscript.com/show_bug.cgi?id=707074.

JorjMcKie:
The problem has been fixed in MuPDF.
The file is damaged in the sense that it contains circular references within its objects - that hadn't been detected before.
Now the circular references are detected and your script with raise an exception - as it should.

julian-smith-artifex-com:
[...]

PR #2718, opened 2023-10-05
fixing #2548

JorjMcKie:
Adding test script


tests/test_2548.py
```diff
@@ -0,0 +1,14 @@
+def test_2548():
```
julian-smith-artifex-com, line 1: We need `import fitz` here.

PR #2720, opened 2023-10-05
Support checking valid PDF name objects

JorjMcKie:
Defines a Python set of characters that are illegal in PDF names. We are currently checking validity of font reference names.

### DIFF
--- a/tests/test_font.py
+++ b/tests/test_font.py
@@ -29,10 +29,10 @@ def test_fontname():
     """Assert a valid PDF fontname."""
     doc = fitz.open()
     page = doc.new_page()
-    assert page.insert_font()
+    assert page.insert_font()  # assert: a valid fontname works!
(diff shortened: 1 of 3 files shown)


tests/test_font.py
```diff
@@ -22,4 +22,18 @@ def test_font2():
...
+    """Assert a valid PDF fontname."""
+    doc = fitz.open()
+    page = doc.new_page()
+    font = fitz.Font("helv")
+    assert page.insert_font()
+    detected = False  # preset indicator
+    try:
+        page.insert_font(fontname="illegal/char", fontbuffer=font.buffer)
+    except ValueError:
+        detected = True  # illegal fontname detected
```
julian-smith-artifex-com, line 38: Might be good to check the text of the exception? Maybe not worth it though, will leave it to you.

PR #2725, opened 2023-10-08
Set key name checks

JorjMcKie:
This change uses the new PDF name verification to confirm that exclusively valid PDF names are specified as parameters in`Document.xref_set_key()` for the key and the value parameter.


julian-smith-artifex-com (approved):

JorjMcKie:
Thank you!

PR #2731, opened 2023-10-10
Japanese Documentation: Part 2

jamie-lemon:
The second phase to complete the translation.


jamie-lemon:
Let's get this in, then we are in a good place to have both docs localizations on `main`! :)

julian-smith-artifex-com (approved):

PR #2700, opened 2023-09-29
Improving Japanese document for better translations and Fixing typos in document

mackey0225:
The other day, I was thrilled to see the Japanese version of the document being provided and made public.
However, I noticed a minor issue in the translation that bothered me, so I made some corrections.

Specifically, "joining and merging" had been translated as "結合と結合," which was a repetition of the same word. I made a slight adjustment to address this.

Furthermore, while reviewing the document, I found several typos and made corrections accordingly.

PR #2737, opened 2023-10-12
Fix #2700

mackey0225:
I apologize for the delay in re-creating the PR. 
I made a mistake with the account, which added some complexity, so I started over from the fork. 
By the way, I haven't included the .po file issue that you had commented on in the revision.


jamie-lemon:
LGTM 👍

PR #2772, opened 2023-10-28
Adding PDF Names resolution

### LINKED ISSUE #2771: Cannot insert text with floating point fontsize/non-zero render mode
Hello,

I recently upgraded from PyMuPDF version 1.22.3 to version 1.23.5 and I can no longer insert text into pages using a floating point fontsize with a render_mode besides zero. All documentation suggests that floats should still be supported, and this approach worked in version 1.22.3 without issue. The issue goes away if I either use a render_mode of 0 or an integer fontsize.

## Describe the bug (mandatory)
PyMuPDF raises the following error when trying to call page.insert_text() with a floating point fontsize and a non-zero render_mode.

image

## To Reproduce (mandatory)
Please use attached script and sample PDF. I believe the same error will be raised for any input PDF though.
[sample.pdf](https://github.com/pymupdf/PyMuPDF/files/13181452/sample.pdf)

[script.py.txt](https://github.com/pymupdf/PyMuPDF/files/13181454/script.py.txt)

## Expected behavior (optional)
Text is inserted without error.

## Your configuration (mandatory)
 - Ubuntu 22.04
 - Python 3.10.6
 - PyMuPDF versions 1.22.3 and 1.23.5, both installed with pip

JorjMcKie:
Add Document method "resolve_names()".
It walks through the "/Dests" and "/Names/Dests" dictionaries in a PDF's catalog and returns a Python dictionary with the name as key and value dictionary containing the target page and location. Note that this method is only implemented in "rebased" and will not work in classic architecture.

Also fixes #2771:
Put expression "border_width * fontsize" in brackets. This is required to correctly set the border width for text output with positive (non-standard) text rendering mode.


src/__init__.py
```diff
@@ -5096,6 +5096,133 @@ def resolve_link(self, uri=None, chapters=0):
...
+        page_xrefs = {self.page_xref(i): i for i in range(self.page_count)}
+
+        def obj_string(obj):
+            """Return string version of a PDF object definition."""
+            buffer = fitz.mupdf.fz_new_buffer(512)
+            output = fitz.mupdf.ll_fz_new_output_with_buffer(buffer.m_internal)
+            fitz.mupdf.ll_pdf_print_obj(output, obj.m_internal, 1, 0)
+            printed = fitz.JM_UnicodeFromBuffer(buffer)
+            fitz.mupdf.ll_fz_drop_output(output)
+            fitz.mupdf.ll_fz_drop_buffer(buffer.m_internal)
```
julian-smith-artifex-com, line 5138: Looks like this code i using the low-level `ll_*` interface because there is no `mupdf.fz_new_output_with_buffer()` - see the comment in mupdf/platform/c++/include/mupdf/classes2.h:
```
	/* Class-aware wrapper for `fz_new_output_with_buffer()`
	is not available because returned wrapper class for `fz_output`
	is non-copyable. */
```
However one can instead construct a `mupdf.FzOutput` directly with `output = fitz.mupdf.FzOutput(buffer)`, and then use `fitz.mupdf.pdf_print_obj()` etc, and not have to worry about calling `fitz.mupdf.ll_fz_drop_output()`.

julian-smith-artifex-com, line 5138: I think this call of `fitz.mupdf.ll_fz_drop_buffer()` might lead to a double-free - `buffer`'s destructor will already call `ll_fz_drop_buffer()`.

JorjMcKie: Ok, cool so far. But how would I read the content of that FzOutput into a Python string (using UTF8)?

tests/test_general.py
```diff
@@ -620,3 +621,16 @@ def test_2635():
...
     pix2 = page.get_pixmap()  # pixmap after cleaning
     assert pix1.samples == pix2.samples  # assert equality
+
+
+def test_resolve_names():
+    """Test PDF name resolution."""
+    # guard against wrong PyMuPDF architecture version
+    if not hasattr(fitz.Document, "resolve_names"):
+        print("PyMuPDF version does not support resolving PDF names")
+        return
```
julian-smith-artifex-com, line 631: [This is not a problem with this PR.] Might be slightly better to explicitly test PyMuPDF version here, just in case `resolve_names()` is accidently deleted or renamed. We have a `mupdf_version_tuple`, i'll look into adding a `pymupdf_version_tuple` or similar.

julian-smith-artifex-com (approved):

PR #2826, opened 2023-11-21
keyword args and flake8 testing and fixes

julian-smith-artifex-com:


JorjMcKie (changes requested):
The change in line 133 of utils.py incorrectly translates "== None".

As per the flake8 test:
We are disabling 20 checks / accepting 20 deviations from a PEP8 compliance.
I disagree with several of them! At least when we are switching over to rebased as the primary implementation, we should not accept deviations that have no better reason than personal preferences.

PR #2845, opened 2023-11-28
Fixed the bug where `button_states()` returns None when `/AP` points to an indirect object.

airgalss:
Description: Some PDF widget objects may not have their `/AP`(Appearance Dictionary) directly pointing to a dictionary, which is enclosed with '<< >>', but rather pointing to an indirect object, expressed by an xref number in the format 'NNN 0 R'. Therefore, the current button_states() only can correctly handle cases where `/AP` points to a dictionary and cannot handle cases where it points to an indirect object. Consequently, I have introduced additional logic to handle the latter scenario, ensuring that `button_states()` can accurately return the on/off state names for button widgets even when `/AP` points to an indirect object.

Test: I have tested the modified `button_states()` function on the mentioned type of PDF, and it now correctly returns the states instead of None.

### DIFF
--- a/fitz/helper-fields.i
+++ b/fitz/helper-fields.i
@@ -1097,10 +1097,26 @@ class Widget(object):
             for x in apnt:
                 nstates.append(x.split()[0])
             states["normal"] = nstates
+        if APN[0] == "xref":
+            nstates = []
+            nxref = int(APN[1].split(" ")[0])
+            APN = doc.xref_object(nxref)
+            apnt = APN.split("/")[1:]
+            for x in apnt:
+                nstates.append(x.split()[0])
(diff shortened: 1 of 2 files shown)


airgalss:

airgalss:
recheck

JorjMcKie:
> recheck

There are lines in `def butten_states()` method that use tab characters instead of 4 spaces.
I would simply change that and copy the method's code to the same-named method in `src/__init__.py`.

airgalss:
I have addressed the issue of mixing tabs and spaces, and copied it to `src/__init__.py`. I hope it should be fine now.

JorjMcKie (approved):

PR #2854, opened 2023-11-30
Add new Page method insert_htmlbox

JorjMcKie:
Note: rebased architecture only.


JorjMcKie:
@julian-smith-artifex-com I am still considering how this new method could be tested.
This will be in a separate PR.

PR #2863, opened 2023-12-03
New Page method "insert_htmlbox"

JorjMcKie:
Add new Page method that inserts HTML/CSS-styled text in a rectangle.

### DIFF
--- a/fitz/utils.py
+++ b/fitz/utils.py
@@ -1834,6 +1834,141 @@ def insert_text(
     return rc
 
 
+def insert_htmlbox(
+    page,
+    rect,
+    text,
+    *,
+    css=None,
+    adjust=True,
+    archive=None,
+    rotate=0,
+    oc=0,
+    overlay=True,
+    morph=None,
+) -> float:
+    """Insert text with optional HTML tags and stylings into a rectangle.
+
+    Args:
+        rect: (rect-like) rectangle into which the text should be placed.
+        text: (str) text with optional HTML tags and stylings.
+        css: (str) CSS styling commands.
+        adjust: (bool) force-fit the text by scaling down (default True).
+        archive: Archive object pointing to locations of used fonts or images
+        rotate: (int) rotate the text with the box by a multiple of 90 degrees.
+        oc: (int) the xref of an OCG / OCMD (Optional Content).
+        overlay: (bool) put text on top of page content.
+        morph: (sequence) (point, matrix) transform result with a matrix
+               using a fixed point.
+    Returns:
+        A float indicating the unused part of the rectangle. If negative, the
+        rectangle was too small for the text.
+    """
+
+    def story_maker(clip, text, adjust=True, rotate=0, css=css, archive=archive):
+        links = []
+
+        def collect(elpos):
+            if not elpos.open_close & 2 or elpos.href is None:
+                return
+            links.append((elpos.href, elpos.rect))
+            return
+
+        rect = Rect(clip)  # copy to rect
+        # shift to top-left = (0, 0)
+        rect += (-clip.x0, -clip.y0, -clip.x0, -clip.y0)
+        if rotate in (90, 270):  # swap width height
+            rect.x1, rect.y1 = rect.y1, rect.x1
+        orig_height = rect.y1  # original relevant height
+        mycss = "body {margin:1px;}" + css  # append user CSS
+        if isinstance(text, str):  # if a string, convert to a Story
+            story = Story(html=text, user_css=mycss, archive=archive)
+        else:
+            story = text  # assume we have a story
+
+        # ---------------------------------------------------------------------
+        # Loop for fitting text into rectangle.
+        # We will expand the rectangle until we have a fit in any case. Only
+        # finally we will check "adjust" and report the overflow amount if
+        # no adjustment was intended.
+        # ---------------------------------------------------------------------
+        more = True
+        while more:
+            story.reset()  # rewind story to the beginning
+            fp = io.BytesIO()  # use for file output
+            writer = DocumentWriter(fp)
+            dev = writer.begin_page(rect)  # make a temporary page
+            more, filled = story.place(rect)  # position text in the rect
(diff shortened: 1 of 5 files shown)


fitz/utils.py
```diff
@@ -1834,6 +1834,141 @@ def insert_text(
...
+        while more:
+            story.reset()  # rewind story to the beginning
+            fp = io.BytesIO()  # use for file output
+            writer = DocumentWriter(fp)
+            dev = writer.begin_page(rect)  # make a temporary page
+            more, filled = story.place(rect)  # position text in the rect
+            if more:  # text did not fit in this clip, so enlarge and try again
+                rect *= 1.01  # enlarge by 1%
+                links = []
+                continue
```
julian-smith-artifex-com, line 1905: If `adjust` is false, should we still be enlarging the rect here?

fitz/utils.py
```diff
@@ -1834,6 +1834,141 @@ def insert_text(
...
+
+    # layout text in rectangle of a temp PDF
+    doc, height, matrix, links = story_maker(
+        rect, text, adjust=adjust, rotate=rotate, css=css, archive=archive
+    )
+
+    rc = rect.height - height if rotate in (0, 180) else rect.width - height
+    if doc is not None:
+        if rc < 0:  # layout was successful: negative values can be ignored
+            rc = 0
```
julian-smith-artifex-com, line 1953: I'm a bit confused by this, and not sure this matches the docs which say that we return negative if unsuccessful, positive if successful.

`rc < 0` means that `rect.height - height < 0`, so `rect.height < height`, so `rect` is not big enough, contrary to the comment `layour was succesful`?

So we seem to be setting `rc` to zero if the layout failed, else positive if layout succeeded?

fitz/utils.py
JorjMcKie: Yes - to enable us computing the (approximate) overflow value.

tests/test_textbox.py
```diff
@@ -144,3 +144,63 @@ def test_2637():
...
+
+    text = """Lorem ipsum dolor sit amet, consectetur adipisici elit, sed eiusmod tempor incidunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation <b>ullamco</b> <i>laboris</i> nisi ut aliquid ex ea commodi consequat. Quis aute iure reprehenderit in <span style="color: #0f0;font-weight:bold;">voluptate</span> velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint obcaecat cupiditat non proident, sunt in culpa qui <a href="https://www.artifex.com">officia</a> deserunt mollit anim id est laborum."""
+
+    doc = fitz.Document()
+
+    for rot in (0, 90, 180, 270):
+        wdirs = ((1, 0), (0, -1), (-1, 0), (0, 1))  # all writing directions
+        page = doc.new_page()
+        page.insert_htmlbox(rect, text, rotate=rot)
```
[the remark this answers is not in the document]
JorjMcKie: The new test script deliberately now uses a a rectangle that is too small. We now try to insert with adjust=False, assert a negative return, then another insert with adjust=True, and assert a non-negative return code.

src/utils.py
```diff
@@ -1872,6 +1872,141 @@ def insert_text(
...
+            writer.end_page()  # end the page
+            writer.close()  # close the writer
+            break  # leave the loop
+
+        try:
+            doc = fitz.open("pdf", fp)  # try make a PDF from memory
+        except Exception as e:
+            # if Document creation failed, return unsuccessful insert
+            print(e)
+            return None, fitz.FZ_MAX_INF_RECT, fitz.Identity, []
```
julian-smith-artifex-com, line 1957: Could we remove the `try..catch`` here and simply let any exception be handled by our caller?

src/__init__.py
```diff
@@ -11939,6 +11939,92 @@ def place( self, where):
...
+                if more:  # did not fit: increase lower factor
+                    low = factor
+                else:
+                    high = factor  # had a fit: decrease upper factor
+                # low / high close enough and we had a fit: exit
+                if high - low < DELTA and not more:
+                    break
+            return 0, factor
+
+        (more, filled) = place(1)
```
julian-smith-artifex-com, line 12026: Minor: Is this equivalent to `self.place(1)`? If so, would prefer to be explicit.

JorjMcKie: It is the same as `self.place(rect)`. Nothing but the effort to ease understanding of the code.

src/__init__.py
```diff
@@ -11939,6 +11939,92 @@ def place( self, where):
...
+
+            if shrink is True:  # decrease rectangle
+                return scale_up_down(LIMIT_L, LIMIT_0)
+            
+            return rect.y1 - filled[3], 1  # rectangle large enough: done
+
+        elif not expand:  # no fit and no scaling
+            return None, -1
+
+        return scale_up_down(LIMIT_0,LIMIT_H)
```
julian-smith-artifex-com, line 12037: This all seems rather complicated to me. And the hard-coded LIMIT_L and LIMIT_H values are a worry.

It looks like we're doing some sort of exponential search when we do `high *= high` or `low *= low` - is that right?

Could we instead make  `insert_htmlbox()` take `scale_min` and `scale_max` args instead of the boolean `scale` arg, and do a conventional binary search?

If we want to be able to allow open-ended max values, we could treat scale_max=None specially and do a simple preliminary loop to find a scaling that is larger than required.

src/utils.py
```diff
@@ -1872,6 +1872,134 @@ def insert_text(
...
+        writer = fitz.DocumentWriter(fp)
+        dev = writer.begin_page(rect * factor)  # make a temporary page
+
+        # get link positions
+        story.element_positions(collect, args={})
+        story.draw(dev)
+        writer.end_page()  # end the page
+        writer.close()  # close the writer
+
+        doc = fitz.Document("pdf", fp)  # reopen memory PDF
```
julian-smith-artifex-com, line 1962: Can we instead use `story.write_with_links()` here?

Issue #2886, opened 2023-12-13
Error in Skeleton for Named Link Destinations


JorjMcKie:
_**Please provide all mandatory information!**_

## Describe the bug (mandatory)
Named link destinations do not work after update/insert.

## To Reproduce (mandatory)
Obvious: even when specifying a correct, existing name in the PDF and preparing a correct link dictionary, the link is not working.

## Expected behavior (optional)
Describe what you expected to happen (if not obvious).

## Your configuration (mandatory)
All configurations

julian-smith-artifex-com:
Fixed in PyMuPDF-1.23.8.

PR #2887, opened 2023-12-13
Fixing #2886

JorjMcKie:
Note: The test script will fail for pymupdf classic in line 24.

### DIFF
--- a/tests/test_2886.py
+++ b/tests/test_2886.py
@@ -1,0 +1,27 @@
+import fitz
+
+import os
+
+
+def test_2886():
+    """Confirm correct insertion of a 'named' link."""
+    path = os.path.abspath(f"{__file__}/../../tests/resources/cython.pdf")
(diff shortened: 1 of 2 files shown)


tests/test_2886.py
```diff
@@ -0,0 +1,27 @@
+import fitz
+
+import os
+
+
+def test_2886():
+    """Confirm correct insertion of a 'named' link."""
+    path = os.path.abspath(f"{__file__}/../../tests/resources/cython.pdf")
```
julian-smith-artifex-com, line 8: Suggest disabling this test when run with classic with:
```
def test_2886():
    if not hasattr(fitz, 'mupdf'):
        print(f'test_2886(): not running on classic.')
        return
    ...
```

JorjMcKie: great - will do

julian-smith-artifex-com (approved):

PR #2891, opened 2023-12-14
Add Page method "insert_htmlbox"

JorjMcKie:
Note:
I removed previous versions because of other fundamental changes.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -1878,6 +1878,125 @@ def insert_text(
     return rc
 
 
+def insert_htmlbox(
+    page,
+    rect,
+    text,
+    *,
+    css=None,
+    scale_low=0,
+    archive=None,
+    rotate=0,
+    oc=0,
+    overlay=True,
+) -> float:
+    """Insert text with optional HTML tags and stylings into a rectangle.
+
+    Args:
+        rect: (rect-like) rectangle into which the text should be placed.
+        text: (str) text with optional HTML tags and stylings.
+        css: (str) CSS styling commands.
+        scale_low: (float) force-fit content by scaling it down. Must be in
+            range [0, 1]. If 1, no scaling will take place. If 0, arbitrary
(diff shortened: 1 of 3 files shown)


JorjMcKie:
Made requested changes.
Decided to after all go away from `scale=True` to a float scale_low, which accepts values inside interval [0,1].
1 means do not scale, 0 is unlimited scale, values in between correspond to scale_max values in `Story.fit_scale()`.

src/utils.py
```diff
@@ -1943,23 +1946,25 @@ def insert_htmlbox(
...
     # ----------------------------------------------------------------
-    if scale is True:
-        scale_max = None
-    else:
+    if scale_low == 1:
         scale_max = 1
+    elif scale_low > 0:
+        scale_max = 1 / scale_low
+    else:
+        scale_max = None
```
julian-smith-artifex-com, line 1954: Can this be `scale_max = None if scale_low == 0 else 1 / scale_low` ?

src/utils.py
```diff
@@ -1943,23 +1946,25 @@ def insert_htmlbox(
...
 
     fit = story.fit_scale(temp_rect, scale_min=1, scale_max=scale_max)
 
     if fit.big_enough is False:  # there was no fit
-        return (-1, 0)
+        return (-1, scale_low)
```
julian-smith-artifex-com, line 1959: Not 100% sure, but i thought the docs said we would return (-1, 0) in this case?

src/utils.py
JorjMcKie: sure

src/utils.py
JorjMcKie: I changed this accordingly

julian-smith-artifex-com:
test_quick is failing because tests do `rc, factor = page.insert_htmlbox(rect, text, rotate=rot, scale=False)`, and  i think `scale=False` now needs to be `scale_low=1`.

PR #2900, opened 2023-12-16
Revive Document.subset_fonts

JorjMcKie:
We were assuring that the generated subset font is not empty by counting its Unicodes - which is not yet implemented in rebased. This fix checks the glyph count instead.


JorjMcKie:
I don't know what has been going wrong. It definitely is not the changes made by this fix.

julian-smith-artifex-com:
> I don't know what has been going wrong. It definitely is not the changes made by this fix.

A commit to MuPDF master yesterday has broken 4 PyMuPDF tests.

julian-smith-artifex-com:
Have checked this PR locally and tests pass if bad mupdf commit is reverted, so will approve now.

julian-smith-artifex-com (approved):

PR #2938, opened 2023-12-22
Adding an action that does a daily build of pyodide wheel.

julian-smith-artifex-com:


julian-smith-artifex-com:
Merging this now; i don't think it needs a review.

PR #3002, opened 2024-01-09
test 2957

JorjMcKie:
Test fix of 2957.

We are using the problem file of issue 2957 and confirm in a hardcoded way that the left over strings in the line do not change positions when applying the redactions.


julian-smith-artifex-com:
I've just updated this PR's branch with a recent change on main, that is needed to make things work with latest mupdf master (on which there's been a change to use of SONAME on Linux). Hopefully it'll fix test_quick, and we can merge the PR.

julian-smith-artifex-com (approved):

PR #3056, opened 2024-01-17
Documentation: Updates feature matrix with item about text writing.

jamie-lemon:
As discussed explains this key feature in the feature matrix with links to relevant API.


JorjMcKie (changes requested):
We should also mention Page.insert_htmlbox (href="page.html#Page.insert_htmlbox") and Page.insert_textbox (href="page.html#Page.insert_textbox").

JorjMcKie (approved):

PR #3074, opened 2024-01-22
Jules 1.23.17

julian-smith-artifex-com:
[This isn't final update for 1.23.17.]


src/__init__.py
JorjMcKie, line 1: What is the instruction in line 5150 doing?
Just a standalone string definition ...

julian-smith-artifex-com: Thanks, yes, that line should not be there. Have force-pushed a fix.

JorjMcKie (approved):

PR #3097, opened 2024-01-25
Things for release of 1.23.19

julian-smith-artifex-com:


tests/test_insertimage.py
JorjMcKie, line 1: shouldn't the mask be xref 6 - not again xref 5?

julian-smith-artifex-com: Thanks, i've added a repeat of the test, using xref 6 for the mask.

PR #3106, opened 2024-01-28
Resolve get_toc speed issue

JorjMcKie:
When executing "get_toc(False)", method "resolve_names()" is executed every time again for each named TOC item. This will lead to intolerable response times.
This hot fix stores the dictionary of resolved names in the Document object - and returns this on subsequent invocations.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -5170,6 +5170,8 @@ class Document:
 
             '21154a7c20684ceb91f9c9adc3b677c40': {'page': -1, 'dest': '/XYZ 15.75 1486 0'}, ...
         """
+        if hasattr(self, "_resolved_names"):  # do not execute multiple times!
+            return self._resolved_names
         # this is a backward listing of page xref to page number
         page_xrefs = {self.page_xref(i): i for i in range(self.page_count)}
 
@@ -5262,6 +5264,7 @@ class Document:
         if tree.pdf_is_dict():
             fill_dict(dest_dict, tree)
 
+        self._resolved_names = dest_dict  # store result or reuse
         return dest_dict
 
     def save(


julian-smith-artifex-com (commented):
Do we need to clear the cache if/when the Document is modified? Otherwise we could end up returning incorrect information.

Minor point - it might be better to set `self._resolved_names = None` in the constructor so we can do `if self.__resolved_names is None: ...` instead of using `hasattr()`. Otherwise it'll get difficult to know what members variables are available.

JorjMcKie:
> Do we need to clear the cache if/when the Document is modified? Otherwise we could end up returning incorrect information.
> 
> Minor point - it might be better to set `self._resolved_names = None` in the constructor so we can do `if self.__resolved_names is None: ...` instead of using `hasattr()`. Otherwise it'll get difficult to know what members variables are available.

I though of this potential problem, too.
In PyMuPDF, there is no way to change the PDF's name destinations. I therefore treat them implicitly as being a constant property of the Document with this fix.
I cannot think of any feature that we do support which would invalidate the PDF's name tree. Should we ever come across one, all that needs to be done is `del doc._resolved_names`. The next invocation of `doc.get_toc(False)` will then automatically re-execute `doc.resolve_names()` fully.

If our existing PyMuPDF features were capable of (implicitely) invalidating the PDF's name tree, we would have seen this issue already. So I am fairly confident we are on the safe side here.

Alternative approach:
Change `Document.get_toc()` such that it invokes `doc.resolve_names()` if the argument is `False`. Then pass that dictionary downstream as an optional parameter.

One way or another: we **_must_** do it. In case of a file like `pandas.pdf`, having 540 NAMED TOC items it is a question of a response time of 11 ms versus minutes.

julian-smith-artifex-com (approved):

JorjMcKie:
Thanks!

PR #3115, opened 2024-01-30
Documentation: Updates header-404.rst to use absolute path for search query.

jamie-lemon:
If I do this: https://pymupdf.readthedocs.io/en/1.22.0/about.html and then I think - "okay i'll use Search" then once I use it it will resolve to e.g.: https://pymupdf.readthedocs.io/en/1.22.0/search.html?q=text+extraction&check_keywords=yes&area=default which is also another 404 page as it is stuck on looking in the 1.22.0 path in this case.

This PR fixes the header-404 search to use an absolute path thus resolving the problem.

### DIFF
--- a/docs/header-404.rst
+++ b/docs/header-404.rst
@@ -8,11 +8,9 @@
 
 
     <div style="display:flex;justify-content:space-between;align-items: center;">
-        <form class="sidebar-search-container top" method="get" action="search.html" role="search" style="width:100%">
+        <form class="sidebar-search-container top" method="get" action="https://pymupdf.readthedocs.io/en/latest/search.html" role="search" style="width:100%">
           <input class="sidebar-search" placeholder="Search" name="q" aria-label="Search">
           <input type="hidden" name="check_keywords" value="yes">
           <input type="hidden" name="area" value="default">
         </form>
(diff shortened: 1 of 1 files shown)


JorjMcKie (approved):

julian-smith-artifex-com (commented):
Could we use `"action=/en/latest/search.html"` here, so that things will work on any server on which the docs are installed (as long as it has the same `en/latest/` hierachy.

Even better, if it works, would be to use a relative path, `"action=../search.html"`. Might have to experiment just in case we need `../../` maybe.

jamie-lemon:
Yes - I guess: `../latest/search.html` would work better here?

jamie-lemon:
Oh wait - we have 2 `404` scenarios:
1) someone searching on a redundant version branch, e.g. https://pymupdf.readthedocs.io/en/1.22.0/about.html 
2) someone with a really bad URL e.g. https://pymupdf.readthedocs.io/en/latest/null_folder/about.html

In the first case asking the search to go up one folder and target `latest` should work, in the second case it wouldn't, I think we will have to use `"action=/en/latest/search.html"` as you initially suggested.

PR #3147, opened 2024-02-12
Multiple fixes

JorjMcKie:
The following is getting fixed:

* remove using fz_image_size() which always has been unreliable based on information of the MuPDF team. We now use the the size of the image's compressed buffer, and, if not present width * height * n of the associated Pixmap.

* issue 3143:
   - use function PyUnicode_FromString() instead of PyUnicode_DecodeRawUnicodeEscape() for layer names
   - return empty string "" if no layer name is present (instead of None).

add flexibility to pixmap-copy

When making a scaled copy of a Pixmap, the last parameter (clip) should be optional.


JorjMcKie:
Hi Julian, this addresses 3 thigs:
* removal of using MuPDF fz_image_size()
* optional 4th parameter (clip) of scaled Pixmap copy (enhancement request 3139)
* fix issue 3143: Turns out that simply using PyUnicode_FromString is the correct Python function - not "raw unicode escape". This fix also has a test script.

src/__init__.py
```diff
@@ -12586,13 +12596,20 @@ def extractIMGINFO(self, hashes=0):
...
+                compr_buff = None
             if hashes:
                 r = mupdf.FzIrect(FZ_MIN_INF_RECT, FZ_MIN_INF_RECT, FZ_MAX_INF_RECT, FZ_MAX_INF_RECT)
                 assert mupdf.fz_is_infinite_irect(r)
                 m = mupdf.FzMatrix(img.w(), 0, 0, img.h(), 0, 0)
                 pix, w, h = mupdf.fz_get_pixmap_from_image(img, r, m)
                 digest = mupdf.fz_md5_pixmap2(pix)
                 digest = bytes(digest)
+                if img_size == 0:
+                    img_size = img.w() * img.h() * pix.n
```
julian-smith-artifex-com, line 12612: This is failing tests. Need `pic.n()`?

JorjMcKie: Changed "pix.n" to using "img.n()"

julian-smith-artifex-com (approved):

PR #3159, opened 2024-02-14
Re-Implement select method

JorjMcKie:
This re-implements Document method "select()" based on new MuPDF function "pdf_rearrange_pages()". This is a more complete (and faster) implementation of what needs to be done here in that not only pages will be rearranged, but also consequential changes will be made to the table of contents, links to removed pages and affected entries in the Optional Content definitions.


src/__init__.py
```diff
@@ -5402,22 +5402,27 @@ def saveIncr(self):
...
+
+        # get underlying pdf document,
         pdf = _as_pdf_document(self)
-        # call retainpages (code copy of fz_clean_file.c)
-        retainpages(pdf, pyliste)
-        if pdf.m_internal.rev_page_map:
-            mupdf.ll_pdf_drop_page_tree(pdf.m_internal)
+
+        # create page sub-pdf via extra.rearrange_pages2
+        extra.rearrange_pages2(pdf, tuple(pyliste))
```
julian-smith-artifex-com, line 5423: We could do:
```
if mupdf_version_tuple >= (1, 24):
    mupdf.pdf_rearrange_pages2(pdf, pyliste)
else:
    extra.rearrange_pages2(pdf, tuple(pyliste))
```
so that we check use of `mupdf.pdf_rearrange_pages2()` when testing with mupdf master. And also it makes it clear that we can remove extra.rearrange_pages2() when we move to a mupdf that provides `mupdf.pdf_rearrange_pages2()`.

JorjMcKie:
Thanks Julian!

PR #3166, opened 2024-02-15
Ensure correct extraction of rotated text in tables

JorjMcKie:
This resolves issue 3148.
The test script creates a table with text rotated in all 4 multiples of 90 degrees. Then extracts and confirms consistency.
Notes:
1. Only rotations by 90° multiples are supported
2. Characters and words within a cell are always assumed to have identical rotation
3. Only rotation 0 cells will respect line breaks. Otherwise, `\n` is replaced by space.

Note: Multiple other table supporting packages will fail this test!


src/table.py
```diff
@@ -387,9 +388,20 @@ def merge_chars(self, ordered_chars: list):
...
+        rotation = 0
+        if not upright and matrix[1] < 0:
+            ordered_chars = reversed(ordered_chars)
+            rotation = 270
+
+        if matrix[0] < 0 and matrix[3] < 0:
+            rotation = 180
+        elif matrix[1] > 0:
+            rotation = 90
+
```
julian-smith-artifex-com, line 404: How come we don't check `upright` when setting rotation to 180 or 90?

And i wonder whether this code could be in the form `if ...: rotation=90 elif ...: rotation=180 elif ...: rotation=270`?

tests/test_tables.py
```diff
@@ -214,3 +214,40 @@ def test_add_lines():
...
+            page.draw_rect(cells[j][i])
+            k = (i + j) % 4
+            page.insert_textbox(cells[j][i] + delta, text[k], rotate=degrees[k])
+    # doc.save("multi-degree.pdf")
+    tabs = page.find_tables()
+    tab = tabs[0]
+    for extract in tab.extract():
+        for item in extract:
+            item = item.replace("\n", " ")
+            assert item in text
```
julian-smith-artifex-com, line 244: Might be good to check that we find each of the items in `text`. E.g.:
```
text_set = set(text)
...
text_set.remove(item)
```

[It looks like `tab.extract()` just returns the text; could it also return the rotation (0, 90, 180, 270)? Then we could additionally check that the rotation matches the text.]

JorjMcKie: No, the format of `table.extract()` is fixed for the time being.

The test that the extracted text of each cell fully matches the multi-word original content already proves that I got everything. If anything would go wrong with a non-0 rotation, the character and / or word sequence would be spoiled - and I am excluding this with that test: IAW the fact that text "rotation by 270 degrees" is extracted literally like this, already is **sufficient proof** that extraction of this angle did work.

If we want to be overly pedantic, we might use a running index and in addition assert the right item in the original array `text`- which has been cyclically reused to originally fill the table.

> I purposefully chose a column count that is not a multiple of 4 to ensure that each rotation occurs in every column at least once.

But that would test a different thing - outside the test scope here, namely that all text items have been extracted in the same sequence as originally inserted (as opposed to only testing that we don't have problems with any of the 4 angles). This

My goal is to ultimately only use our own text extraction and especially text / word re-composition. That would allow for fully reliably extracting all text meta-information (color, font, font size, etc.)

src/table.py
JorjMcKie: The "upright" variable is sort of a reserved word in the original table extraction logic (pdfplumber). It always means 90° if set. I won't tear apart everything in the logic to make that more meaningful.
pdfplumber ignores any rotation angles other than 0 and 90. It is also incapable of correctly extracting multi-line content if rotation is not 0. But we now can do it.
As long as I am bound to use pdfplumber's own word extraction (and not PyMuPDF's more elegant and faster methods) I need to use other indicators for dealing with things outside pdfplumber's reach.

julian-smith-artifex-com (approved):

PR #3173, opened 2024-02-16
pipcl.py: allow providing python-config externally

rmelotte:
When cross-compiling (e.g. using Buildroot), the python-config executable that resides next to the host python executable provides incorrect includes (the ones for the host).

Since the correct path to python-config cannot be guessed, add an additional environment variable to allow setting the path to the correct python-config executable externally.

### DIFF
--- a/pipcl.py
+++ b/pipcl.py
@@ -1769,6 +1769,9 @@ class PythonFlags:
     Compile/link flags for the current python, for example the include path
     needed to get `Python.h`.
 
+    The 'PIPCL_PYTHON_CONFIG' environment variable allows to override
+    the location of the python-config executable.
+
     Members:
         .includes:
             String containing compiler flags for include paths.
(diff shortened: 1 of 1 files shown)


rmelotte:

julian-smith-artifex-com (commented):
Many thanks for this.

If we are to make pipcl.py use an environment variable, i think it should be `PIPCL_PYTHON_CONFIG`, because `pipcl.py` is (or tries to be) a generic python backend.

[Alternatively, because PyMuPDF's `setup.py` uses various `PYMUPDF_*` environment variables, maybe we should add a new `python_config` arg to `pipcl.py:Package.__init__()`, and set it in `setup.py` from `PYMUPDF_PYTHON_CONFIG`? If you want to keep things simple for this PR though, please don't worry about doing this.]

julian-smith-artifex-com:
[Oh and don't worry about the failure of test_quick - there's a known problem with mupdf master at the moment.]

julian-smith-artifex-com:
BTW i've just rebased this PR's commits on top of recent changes that have fixed test_quick,m so that test_quick passes in the PR. You'll need to `git fetch` and do a hard reset to ensure your local branch matches the branch on GitHub.com. Apologies if this is inconvenient.

rmelotte:
@julian-smith-artifex-com Thanks for your feedback!
To keep it simple for now, I've only renamed the variable to `PIPCL_PYTHON_CONFIG` and updated the documentation.

Thanks also for the rebase, I also forced-pushed it just now to keep it as a single commit.

julian-smith-artifex-com:
Many thanks, i'll merge this now.

Issue #3186, opened 2024-02-20
extractText() extracts broken text from pdf


spagliarini:
### Description of the bug

Hi, 

I noticed a bug in PyMuPDF version > 1.23.9 (included) when using get_text to extract text from PDF documents.

To reproduce the bug 
- Consider the attached PDF file: [test_file.pdf](https://github.com/pymupdf/PyMuPDF/files/14343604/test_file.pdf)

- Extract text using the code below (see "How to reproduce the bug")

- To reproduce the correct behavior install a PyMuPDF version < 1.23.9 (e.g., 1.23.8). We obtain the following **complete** text: [doc_text_1238.txt](https://github.com/pymupdf/PyMuPDF/files/14343570/doc_text_1238.txt)

- To reproduce the bug behavior install a PyMuPDF version >= 1.23.9 (e.g., 1.23.24). We obtain the following **broken** text: [doc_text_12324.txt](https://github.com/pymupdf/PyMuPDF/files/14343674/doc_text_12324.txt)

**ADDITIONAL NOTES**
- The bug behavior can only be observed on certain documents (e.g., the one attached above)
- extractBLOCKS, extractWORDS and extractDICT work fine, the bug seems to show only for extractTEXT
- We tried both on windows and linux and neither works

Thank you for your help 

### How to reproduce the bug

Extract text using the following code
```python
fitz_doc = fitz.open(pdf_path)

doc_text = list()
for page in fitz_doc:
    doc_text.append(page.get_text())

doc_text = ' '.join(doc_text)
```

To reproduce the bug behavior install a PyMuPDF version >= 1.23.9 (e.g., 1.23.24).

### PyMuPDF version

1.23.24

### Operating system

Windows

### Python version

3.10

julian-smith-artifex-com:
Fixed in 1.23.25.

Issue #3182, opened 2024-02-19
Pixmap.invert_irect argument type error


luhuaei:
### Description of the bug
[Pixmap.invert_irect](https://pymupdf.readthedocs.io/en/latest/pixmap.html#Pixmap.invert_irect)
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/site-packages/fitz/__init__.py", line 9841, in invert_irect
    r = JM_irect_from_py(bbox)
        ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/fitz/__init__.py", line 16270, in JM_irect_from_py
    ret = mupdf.fz_make_irect(r.x0, r.y0, r.x1, r.y1)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/mupdf/__init__.py", line 39687, in fz_make_irect
    return _mupdf.fz_make_irect(x0, y0, x1, y1)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: in method 'fz_make_irect', argument 1 of type 'int'
```

### How to reproduce the bug

```
>>> import fitz
>>> pix = fitz.Pixmap("screenshot.png")
>>> rect = fitz.Rect(0, 0, 100, 100)
>>> pix.invert_irect(rect)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/site-packages/fitz/__init__.py", line 9841, in invert_irect
    r = JM_irect_from_py(bbox)
        ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/fitz/__init__.py", line 16270, in JM_irect_from_py
    ret = mupdf.fz_make_irect(r.x0, r.y0, r.x1, r.y1)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/mupdf/__init__.py", line 39687, in fz_make_irect
    return _mupdf.fz_make_irect(x0, y0, x1, y1)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: in method 'fz_make_irect', argument 1 of type 'int'
[...]

PR #3187, opened 2024-02-20
Address #3182: Pixmap.invert_irect argument type error.

julian-smith-artifex-com:
Address #3182 and #3186.


src/__init__.py
```diff
@@ -16267,15 +16267,15 @@ def JM_irect_from_py(r):
         r = mupdf.FzIrect( r.x0, r.y0, r.x1, r.y1)
         return r
     if isinstance(r, Rect):
-        ret = mupdf.fz_make_irect(r.x0, r.y0, r.x1, r.y1)
+        ret = mupdf.fz_make_irect(int(r.x0), int(r.y0), int(r.x1), int(r.y1))
```
JorjMcKie, line 16270: I don't think this is correct, because rounding is different from how MuPDF makes a IRect from a Rect: uppper and left coords must be rounded in direction left / up, but lower / right coords must take the next available integer ceiling.

julian-smith-artifex-com: Thanks, i hadn't realised that. New commit fixes hopefully.

JorjMcKie (approved):

PR #3210, opened 2024-02-28
Immunize against empty table headers

JorjMcKie:
Improved check for empty tables, fixes bugs when determining table headers.

More complete and faster computation of enveloping vector graphic rectangles.


src/table.py
```diff
@@ -1932,11 +1959,22 @@ def are_neighbors(r1, r2):
...
+                or r2.y0 - snap_y <= r1.y1 <= r2.y1 + snap_y
+            )
+            or (
+                r1.x0 - snap_x <= r2.x0 <= r1.x1 + snap_x
+                or r1.x0 - snap_x <= r2.x1 <= r1.x1 + snap_x
+            )
+            and (
+                r1.y0 - snap_y <= r2.y0 <= r1.y1 + snap_y
+                or r1.y0 - snap_y <= r2.y1 <= r1.y1 + snap_y
+            )
```
julian-smith-artifex-com, line 1977: I don't know what it highest priority, `and` or `or`. Could we make this explicit with brackets here? I.e. instead of `if (expr1) and (expr2) or (expr3) and (expr4):`, suggest either `if ((expr1) and (expr2)) or ((expr3) and (expr4)):`, or `if (expr1) and ((expr2) or (expr3)) and (expr4):`.

src/table.py
```diff
@@ -1959,29 +1997,37 @@ def clean_graphics():
         new_rects = []  # the final list of joined rectangles
         # -------------------------------------------------------------------------
         # Strategy: Join rectangles that "almost touch" each other,
-        # Extend first rectangle with any remaining in the list that touches it.
+        # Extend first rectangle with any other that is a "neighbor".
         # Then move it to final list and continue with the rest.
         # -------------------------------------------------------------------------
         while prects:  # the algorithm will empty this list
-            r = prects[0]  # first rectangle
+            r = +prects[0]  # first rectangle
```
julian-smith-artifex-com, line 2004: What does unary `+` do here?

src/table.py
```diff
@@ -1959,29 +1997,37 @@ def clean_graphics():
...
+        return [
+            r
+            for r in new_rects
+            if not white_spaces.issuperset(
+                page.get_textbox(
+                    r,
+                    textpage=TEXTPAGE,
+                )
+            )
+        ], paths
```
julian-smith-artifex-com, line 2030: I had to struggle quite hard before i understood what this list comprehension does. Prefer a loop for clarity. But feel free to ignore as this isn't important.

src/table.py
JorjMcKie: You don't? Well I can of course add more brackets ...

src/table.py
JorjMcKie: This is the fastest version of the logic, because the comprehension is internally compiled - as opposed to an explicit loop.
Especially since the latest Python version, another performance improvement has been implemented by no longer internally creating a function object representing the comprehension.

What I might want to do is instead something entirely different - which will indeed need to be implemented via a separate loop:

What I am doing here is identifying bboxes that wrap separate vector graphics belonging together. The borders of these wrappers are later used like virtual vector graphics themselves thus helping to identify tables and cells.

If in the wrapping bbox is no (real) text, then not only the bbox, but all other vector graphics inside it should be likewise ignored - i.e. deleted from the `paths` list. This should speed up table recognition further down.

Might be worthwhile to do this now - as long as there is no urgency b/o a new release _ante portas._

JorjMcKie: Ok it was easier than I thought: the complex list comprehension is gone with the next update.

julian-smith-artifex-com (approved):

PR #3219, opened 2024-03-02
Fixes

JorjMcKie:
Address bug 3207:
Multiple sub-paths in a path were not correctly handled when "closepath" commands were present. When encountering "closepath", this fix generates a line that connects the current point with the first point of the current sub-path - however only if these two points are different. The "firstpoint" value is being set exclusively by the "moveto" command.


jamie-lemon (commented):
page.rst docs update LGTM

src/__init__.py
```diff
@@ -19282,6 +19299,8 @@ def __init__(self, out, clips, method):
         self.ctm = mupdf.FzMatrix()
         self.rot = mupdf.FzMatrix()
         self.lastpoint = mupdf.FzPoint()
+        self.firstpoint = mupdf.FzPoint()
+        self.havemove = 0
```
julian-smith-artifex-com, line 19303: Minor suggestion - we could set self.firstpoint to None here, and then use `if self.firstpoint is not None: ...` instead of `if self.havemove: ...` later on.

julian-smith-artifex-com (approved):

src/__init__.py
JorjMcKie: how to reflect that in extract.i?

PR #3266, opened 2024-03-14
installation.rst: remove need to install tesseract

deeplow:
As stated in the document already, Tesseract OCR is already bundled into the application.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -206,7 +206,7 @@ Enabling Integrated OCR Support
 
 If you do not intend to use this feature, skip this step. Otherwise, it is required for both installation paths: **from wheels and from sources.**
 
-PyMuPDF will already contain all the logic to support OCR functions. But it additionally does need Tesseract's language support data, so installation of Tesseract-OCR is still required.
+PyMuPDF will already contain all the logic to support OCR functions. But it additionally does need [Tesseract’s language support data](https://github.com/tesseract-ocr/tessdata).
 
 The language support folder location must be communicated either via storing it in the environment variable `"TESSDATA_PREFIX"`, or as a parameter in the applicable functions.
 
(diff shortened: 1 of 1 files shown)


jamie-lemon:
Thanks for pointing this stuff out.

I think for this one it is better if bullet point 1, remains and it says:

`1. Ensure that Tesseract is installed`

Because we can't assume it is installed (even though it probably is vi the MuPDF installation), possibly someone could have removed it - who knows?

Also further up the section it says in the text: "PyMuPDF will already contain all the logic to support OCR functions. But it additionally does need Tesseract’s language support data, so installation of Tesseract-OCR is still required."

Perhaps that needs to say:  "PyMuPDF will already contain all the logic to support OCR functions. But it additionally does need Tesseract’s language support data, which should already be installed on your system."

jamie-lemon:
Fair enough, we can top the 1st bullet point. If you update this PR to do this part further up:
"PyMuPDF will already contain all the logic to support OCR functions. But it additionally does need Tesseract’s language support data, so installation of Tesseract-OCR is still required."
->
"PyMuPDF will already contain all the logic to support OCR functions. But it additionally does need Tesseract’s language support data, which should already be installed on your system."

Then we are good. :)

jamie-lemon:
Sure - we try to avoid links which are not descriptive - so [here](https://github.com/tesseract-ocr/tessdata) is out !

Let's say:

PyMuPDF will already contain all the logic to support OCR functions. But it additionally does need [Tesseract’s language support data](https://github.com/tesseract-ocr/tessdata).

jamie-lemon:
Nearly - because this is RST markup and not straight MD links are a bit different - you need to change it to:

Screenshot 2024-03-22 at 13 49 58

I have to do a screen shot here to prevent this comment form parsing the backticks as code :)

See: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages

PR #3289, opened 2024-03-21
Several fixes for table module

JorjMcKie:
- Add new method for outputting the table as a markdown string.

- Address errors in computing the table header object: We now allow None as the cell value, because this will be resolved where needed (e.g. in the pandas DataFrame).

We previously tried to enforce rect-like tuples in all header cell bboxes, however this fails for tables with all-None columns. This fix enables this and constructs an empty string in the corresponding cell string.

We now correctly include start / stop points of lines in the bbox of the clustered graphic. We previously joined the line's rectangle - which had no effect because this is always empty.

### DIFF
--- a/src/table.py
+++ b/src/table.py
@@ -75,6 +75,7 @@ This is implemented as new class TableHeader with the properties:
 import inspect
 import itertools
 import string
+import html
 from collections.abc import Sequence
 from dataclasses import dataclass
 from operator import itemgetter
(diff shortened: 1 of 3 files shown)


src/table.py
```diff
@@ -1365,6 +1365,46 @@ def char_in_bbox(char, bbox) -> bool:
...
+        output = "|"
+
+        # generate header string and MD underline
+        for i, name in enumerate(self.header.names):
+            if name is None or name == "":  # generate a name if empty
+                name = f"Col{i+1}"
+            name = name.replace("\n", " ")  # remove any line breaks
+            if clean:  # remove sensitive syntax
+                name = (
+                    name.replace("<", "&lt;").replace(">", "&gt;").replace("-", "&#45;")
```
julian-smith-artifex-com, line 1381: `html.escape` might be better here and elsewhere? Also escapes `&`, and doesn't escape `-` which seems better?

JorjMcKie: where do I find definition of `html.escape`?

julian-smith-artifex-com (approved):

PR #3302, opened 2024-03-23
more permissive URI kind links

JorjMcKie:
We previously only accepted external links as LINK_URI if the "uri"-string started with a small list of prefixes. With this change /URI links are turned into LINK_URI links if they contain a colon(":") and do not start with prefix "file://". In all other cases, a LINK_LAUNCH will be generated.


tests/test_toc.py
```diff
@@ -24,7 +25,9 @@ def test_simple_toc():
 
 
 def test_full_toc():
-    if fitz.mupdf_version_tuple >= (1, 23, 0):
+    if fitz.pymupdf_version_tuple >= (1, 24, 0) and hasattr(fitz, "mupdf"):
```
julian-smith-artifex-com, line 28: I'm not sure tests should ever look at `pymupdf_version_tuple`. The only reason for doing so would be to allow a git checkout of PyMuPDF to be used to test an older version of PyMuPDF that has been installed locally. This is hardly ever useful i think, and it introduces a potentially complicated extra set of decisions in the test code.

There are a couple of other places where i've incorrectly written test code that looks at `pymupdf_version_tuple`; i will remove these references.

[Having test code look at fitz.`mupdf_version_tuple` is different i think, because it allows us to test the latest PyMuPDF code with MuPDF release and master branches, both of which are useful things to do.]

src/__init__.py
```diff
@@ -6797,7 +6796,10 @@ def uri_to_dict(uri):
                     if ftab[1].startswith("page="):
                         self.kind = LINK_GOTOR
                         self.file_spec = ftab[0]
-                        self.page = int(ftab[1][5:]) - 1
+                        self.page = int(ftab[1].split("&")[0][5:]) - 1
+            elif ":" in self.uri:
```
julian-smith-artifex-com, line 6800: Could we tighten this to look for a colon only after an initial set of alphabetical characters?

For example:
```
if re.match('^[a-z]+:', self.uri)`:
    ...
```

julian-smith-artifex-com (approved):

PR #3328, opened 2024-04-01
Added pylint test and fixed various issues it detects.

julian-smith-artifex-com:
scripts/gh_release.py
    Add 'pylint' to `test_packages`.
scripts/sysinstall.py
    Exclude pylint in sysinstall tests - pylint seems to get confused and
    doesn't think `message()` is in module `fitz`.
src/__init__.py
    Annot.apn_bbox(): fixed broken call to self.get_parent().derotation_matrix().
    Annot.set_apn_bbox(): fixed broken call to page.rotation_matrix().
    Annot.set_apn_matrix(): added missing `self` arg.
    Annot.set_irt_xref(): added missing `self` arg.
    Archive._add_arch(): added missing `self` arg.
    Document:
        Added properties metadata() and name(), and moved members .metadata and
        .name to ._metadata and ._name; this fixes error from pylint, which
        couldn't see the old .name and .metadata members.
    Don't define top-level names `self`, `name`, `value`, because they cause
    pylint warnings if functions (re)use these names.
src/__main__.py
    Removed unused `Tuple` import.
    Changed `mycentre` from lambda to function.
    Removed superflous trailing `return` statements.
    Changed empty `except:` to `except Exception:`.
    Fixed incorrect specification of `mupdf.FzErrorBase`.
    Fixed incorrect call to doc.saveIncr().
    Use `with open()...` instead of explicit closing.
src/extra.i
    Disable all pylint warnings on SWIG-generated python.
src/fitz.py
    Disable pylint wildcard-import,unused-import warnings otherwise pylint
    complains a lot.
src/table.py
    Removed unnecessary inheritence of `object`.
src/utils.py
    Avoid pylint warnings when accessing `fitz.PDF_ANNOT_LINK` etc, which are
    not defined explicitly in `__init__.py`.
    Disabled various other pylint warnings.
    Fixed bad call to `message()`.
    Use dict.items() instead of explicitly looking up keys.
tests/test_pylint.py
    New, contains test_pylint(). We currently disable many non-critical
    warnings.

### DIFF
--- a/scripts/gh_release.py
+++ b/scripts/gh_release.py
@@ -674,7 +674,7 @@ def platform_tag():
         assert 0, f'Unrecognised: {platform.system()=}'
 
 
-test_packages = 'pytest fontTools pymupdf-fonts flake8'
+test_packages = 'pytest fontTools pymupdf-fonts flake8 pylint'
 if platform.system() == 'Windows' and cpu_bits() == 32:
     # No pillow wheel available, and doesn't build easily.
     pass
--- a/scripts/sysinstall.py
+++ b/scripts/sysinstall.py
@@ -340,7 +340,9 @@ def main():
     run(f'ls -l {root_prefix}/bin/')
     # 2024-03-20: Not sure whether/where `pymupdf` binary is installed, so we
     # disable the test_cli* tests.
-    command += f' pytest -k "not test_color_count and not test_3050 and not test_cli and not test_cli_out" {pymupdf_dir}'
+    excluded_tests = 'test_color_count test_3050 test_cli test_cli_out test_pylint'.split()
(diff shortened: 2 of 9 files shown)


JorjMcKie (approved):
So pylint doesn't like standalone lambda functions, does it.
Well, no strong feelings here.

Other recommendations are very worthwhile to consider, respectively a MUST as part of your grand code overhaul.

The recommendation to use "x = max(x, y)" instead of "x = y if y > x else x" is however a typical Python mannerism: using the max() function in this way has a 4 times longer execution time - a no-go in performance-critical code.

Out of performance arguments we should indeed use list / dict / set comprehensions wherever possible (while keeping the code fairly easy to understand).

PR #3358, opened 2024-04-07
Add Page method "cluster_drawings"

JorjMcKie:
New method "Page.cluster_drawings" identifies drawing items (i.e. vector graphics or line-art) that belong together based on their geometrical vicinity and computes rectangles that wrap such "clusters".

Another change adds a new parameter "keep_text" to `Page.apply_redactions()` to reflect MuPDF's capability to independently keep or remove the three object categories text, images and vector graphics during applying redactions.

### DIFF
--- a/docs/page.rst
+++ b/docs/page.rst
@@ -63,6 +63,7 @@ In a nutshell, this is what you can do with PyMuPDF:
 :meth:`Page.annots`                return a generator over the annots on the page
 :meth:`Page.apply_redactions`      PDF only: process the redactions of the page
 :meth:`Page.bound`                 rectangle of the page
+:meth:`Page.cluster_drawings`      PDF only: bounding boxes of vector graphics
 :meth:`Page.delete_annot`          PDF only: delete an annotation
 :meth:`Page.delete_image`          PDF only: delete an image
 :meth:`Page.delete_link`           PDF only: delete a link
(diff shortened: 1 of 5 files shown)


src/__init__.py
```diff
@@ -8923,6 +8924,101 @@ def get_lineart(self) -> object:
...
+
+            if (
+                r1.x0 - delta_x <= r2.x0 <= r1.x1 + delta_x
+                or r1.x0 - delta_x <= r2.x1 <= r1.x1 + delta_x
+            ) and (
+                r1.y0 - delta_y <= r2.y0 <= r1.y1 + delta_y
+                or r1.y0 - delta_y <= r2.y1 <= r1.y1 + delta_y
+            ):
+                return True
+            return False
```
julian-smith-artifex-com, line 8984: I'm probably missing some subtlety here, but we seem to be assuming that these rects are correctly ordered, so could we simplify this to check for overlap like:
```
    if (1
            or r1.x1 < r2.x0-delta
            or r1.x0 > r2.x1+delta
            or r1.y1 < r2.y0-delta
            or r1.y0 > r2.y1+delta
            ):
        # Rects do not overlap.
        return False
    else:
        # Rects overlap.
        return True
```
?

src/__init__.py
```diff
@@ -8923,6 +8924,101 @@ def get_lineart(self) -> object:
...
+
+        # exclude graphics not contained in the clip
+        paths = [
+            p
+            for p in drawings
+            if parea.x0 <= p["rect"].x0 <= p["rect"].x1 <= parea.x1
+            and parea.y0 <= p["rect"].y0 <= p["rect"].y1 <= parea.y1
+        ]
```
julian-smith-artifex-com, line 8992: Similar to above, i'm again probably missing something, but could we simplify this to:
```
paths = [
        p for p in drawings
        if not (0
                or p.x1 < p['rect'].x0
                or p.x0 > p['rect'].x1
                or p.y1 < p['rect'].y0
                or p.y0 > p['rect'].y1
                )
        ]
```?

File: src/utils.py [hunk 3]
```diff
@@ -4186,11 +4186,14 @@ def commit(self, overlay: bool = True) -> None:
         self.totalcont = ""  # re-use
 
 
-def apply_redactions(page: fitz.Page, images: int = 2, graphics: int = 1) -> bool:
+def apply_redactions(page: fitz.Page, keep_text: int = 0, images: int = 2, graphics: int = 1) -> bool:
```
julian-smith-artifex-com, line 4189: Could we append the new `keep_text` arg to the fns args? Otherwise existing code that specifies things as positional args will break. [We should probably move to using `*` to force naming of args like these.]

The new `keep_text` arg seems to behave in the opposite way to the existing args, which could be confusing. Could we instead call it `text` and use zero to mean "ignore text" and one to mean "remove overlapping text"?

tests/test_annots.py
```diff
@@ -158,8 +159,8 @@ def test_redact1():
         rotate=-1,
     )
     assert annot.type == (12, "Redact")
-    x = annot._get_redact_values()
-    pix = annot.get_pixmap()
+    annot._get_redact_values()
+    annot.get_pixmap()
```
julian-smith-artifex-com, line 163: It's not that important, but this test is calling internal implementation method `annot._get_redact_values()`, which seems a bit unusual. Could the test be changed to only use the public API?

src/__init__.py
JorjMcKie: That's not simpler. I could dissolve my expression into
```python
if (parea.x0 <= p["rect"].x0 and
    p["rect"].x1 <= parea.x1 and
    parea.y0 <= p["rect"].y0 and
    p["rect"].y1 <= parea.y1
    )
```
if you insist. This is also faster because it exits as soon as a condition is not met - in contrast to always evaluating all "or's".

src/utils.py
JorjMcKie: I will change this to using the `*` parameter - good point.
Otherwise, the reason for choosing the name `keep_text` is two-fold:
1. When interpreting 0 as False (as is usual), the condition could sound like "text is false", giving rise to the wrong association text will not be removed.
2. The original intent of redactions is to remove text - for legal / data protection reasons. From this perspective, **_not_** removing text is a paradox desire, which can justify an uncommon argument naming.

docs/page.rst
```diff
@@ -322,16 +323,17 @@ In a nutshell, this is what you can do with PyMuPDF:
...
 
-      :arg int graphics: How to redact overlapping vector graphics (also called "line art" or "drawings"). The default (2) removes any overlapping vector graphics. `PDF_REDACT_LINE_ART_NONE` (0) ignores, and `PDF_REDACT_LINE_ART_IF_COVERED` (1) removes graphics fully contained in a redaction annotation.
+      :arg int graphics: How to redact overlapping vector graphics (also called "line-art" or "drawings"). The default (2) removes any overlapping vector graphics. `PDF_REDACT_LINE_ART_NONE | 0` ignores, and `PDF_REDACT_LINE_ART_IF_COVERED | 1` removes graphics fully contained in a redaction annotation. When removing line-art, please be aware that **stroked** vector graphics (i.e. type "s" or "sf") have a **larger wrapping rectangle** than one might expect: first of all, at least 50% of the path's line width have to be added in each direction to truly include all of the drawing. If a so-called "miter limit" is provided (see page 121 of the PDF specification), the enlarging value is `miter * width / 2`. When letting everything default, the resulting
```
jamie-lemon, line 334: the resulting ..... ?

How should this sentence end?

jamie-lemon (changes requested):

jamie-lemon (approved):

julian-smith-artifex-com (approved):

PR #3401, opened 2024-04-19
Updates CLA assistant branch path.

jamie-lemon:
This ensures that if we have branch protection for our `main` branch turned on which only allowed merges with an associated approved PR then the CLA assistant can still commit  as it targets a separate branch ( the `CLA` branch).


julian-smith-artifex-com:
Before this commit is pushed, presumably we need to:

* Create the new branch `CLA` from current branch main, and `git rm` all files except for `signatures/version1/cla.json`.
* Remove `signatures/version1/cla.json` from the branch `main`.

?

jamie-lemon:
I did create the CLA branch with a simple CLA-README. Never thought about removing everything else from that branch but I guess it makes sense.

jamie-lemon:
Okay - I removed everything else from the CLA branch ( which felt pretty scary, like I'm deleting all the PyMuPDF source code, but I'll get over it :) ) 
Please check , then I guess I could just add the removal of `signatures/version1/cla.json` to this PR ?

julian-smith-artifex-com (approved):

PR #3406, opened 2024-04-23
Add new page method "remove_rotation"

JorjMcKie:
The new method sets a page's rotation to 0 while keeping the page's appearance and content unchanged. This can significantly simplify inclusion of the page in target pages via Page method "show_pdf_page()".

Other changes are additions / corrections to "changes.txt" and improved wordings for in document.rst and page.rst.

Also contained is a fix for issue 3402 (inability to insert/update fields with inter-field calculations.


changes.txt
```diff
@@ -11,6 +28,7 @@ Change Log
 
   * **Fixed** `3331 <https://github.com/pymupdf/PyMuPDF/issues/3331>`_: Document.pages() is incorrectly type-hinted
   * **Fixed** `3354 <https://github.com/pymupdf/PyMuPDF/issues/3354>`_: PyMuPDF==1.24.1: AttributeError: property 'metadata' of 'Document' object has no setter
+  * **Fixed** `3376 <https://github.com/pymupdf/PyMuPDF/issues/3376>`_: PyMuPDF==1.24.1: Redacting results are not as expected in 1.24.1.
```
julian-smith-artifex-com, line 31: I don't think this was fixed in 1.24.2.

src/utils.py
```diff
@@ -480,16 +480,11 @@ def get_text_blocks(
...
     """
     fitz.CheckParent(page)
     if flags is None:
-        flags = (
-                fitz.TEXT_PRESERVE_WHITESPACE
-                    | fitz.TEXT_PRESERVE_IMAGES
-                    | fitz.TEXT_PRESERVE_LIGATURES
-                    | fitz.TEXT_MEDIABOX_CLIP
-                )
+        flags = fitz.TEXTFLAGS_BLOCKS
```
julian-smith-artifex-com, line 487: Could you explain this change in the commit message? It seems to be changing to TEXTFLAGS_BLOCKS=
        | TEXT_PRESERVE_LIGATURES
        | TEXT_PRESERVE_WHITESPACE
        | TEXT_MEDIABOX_CLIP
        | TEXT_CID_FOR_UNKNOWN_UNICODE
?

changes.txt
JorjMcKie: correct - also found that out in the meantime ...

src/utils.py
JorjMcKie: This method is an alias for get_text("blocks") so it must use the same flags. We wouldn't be able to give a good answer when being asked why the behavior is different.

julian-smith-artifex-com (approved):

Issue #3414, opened 2024-04-26
missing attribute set_dpi()


tobiasfarmerenemy:
### Description of the bug

I want to create a pixmap from an annot and adapt the resolution with the "dpi" argument as stated in the docs.
However, this is creating an 

Attribute Error: 'FzPixmap' object has no attribute 'set_dpi'

is this a bug?

### How to reproduce the bug

annot.get_pixmap(dpi=10)

### PyMuPDF version

1.24.2

### Operating system

Windows

### Python version

3.11

julian-smith-artifex-com:
Fixed in 1.24.3.

PR #3417, opened 2024-04-27
set DPI using mupdf attributes

JorjMcKie:
Addresses 3414

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -732,10 +732,11 @@ class Annot:
         if not cs:
             cs = mupdf.fz_device_rgb()
 
-        pix = mupdf.pdf_new_pixmap_from_annot( self.this, ctm, cs, mupdf.FzSeparations(0), alpha)
+        pix = mupdf.pdf_new_pixmap_from_annot(self.this, ctm, cs, mupdf.FzSeparations(0), alpha)
+        ret = Pixmap(pix)
         if dpi:
-            pix.set_dpi(dpi, dpi)
-        return Pixmap( pix)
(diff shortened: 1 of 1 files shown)


src/__init__.py
```diff
@@ -732,10 +732,11 @@ def get_pixmap(self, matrix=None, dpi=None, colorspace=None, alpha=0):
         if not cs:
             cs = mupdf.fz_device_rgb()
 
-        pix = mupdf.pdf_new_pixmap_from_annot( self.this, ctm, cs, mupdf.FzSeparations(0), alpha)
+        pix = mupdf.pdf_new_pixmap_from_annot(self.this, ctm, cs, mupdf.FzSeparations(0), alpha)
         if dpi:
-            pix.set_dpi(dpi, dpi)
-        return Pixmap( pix)
+            pix.m_internal.xres = dpi
+            pix.m_internal.yres = dpi
```
julian-smith-artifex-com, line 738: Why this change - isn't it just duplicating the code in `Pixmap.set_dpi()`?

julian-smith-artifex-com: Oh, i see the reference to #3414. But might it be better to do:
```
ret = Pixmap(pix)
ret.set_dpi(dpi, dpi)
return ret
```
? This way we make use of the existing `Pixmap.set_dpi()`, instead of duplicating it.

JorjMcKie: Ah, ok - didn't know this would work like that.

JorjMcKie: fixed

julian-smith-artifex-com (approved):

PR #3426, opened 2024-04-30
fix test_mupdf_regressions version check for bug 707727

mjg:
Bug 707727 is fixed in mupdf's master branch, not in the release branch and in particular not in 1.24.1. So adjust the check, assuming that the fix will be backported to 1.24.2

Depending on your plans for the backports, the version check needs be adjusted, of course. The proposed change is the minimal change which works as of today.


mjg:

mjg:
(Not that I think that adjusting a version number is copyrightable, but ...)

JorjMcKie (approved):

julian-smith-artifex-com:
@mjg Thanks for taking the trouble to submit this PR. I've just verified that is has fixed the failure of the Github action "Test mupdf release branch".

PR #3437, opened 2024-05-03
Allow `from fitz.utils import *` etc to work.

julian-smith-artifex-com:
We create a `fitz` directory instead of a `fitz.py` file in wheels and when installing. Then `fitz/utils.py` contains `from pymupdf.utils import *` etc.


julian-smith-artifex-com:
[This fixes use with pdf2docx.]

JorjMcKie (approved):

PR #3500, opened 2024-05-17
Fontsubset docs

cbm755:
The return type is different b/w the new way and the deprecated implementation.  Also, `verbose` only does something for the old code.  `fontTools` is needed only for the deprecated implementation.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -5360,7 +5360,7 @@ def recover_char_quad(line_dir: tuple, span: dict, char: dict) -> pymupdf.Quad:
 # Building font subsets using fontTools
 # -------------------------------------------------------------------
 def subset_fonts(doc: pymupdf.Document, verbose: bool = False, fallback: bool = False) -> None:
-    """Build font subsets of a PDF. Requires package 'fontTools'.
+    """Build font subsets in a PDF.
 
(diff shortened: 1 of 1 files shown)


src/utils.py
```diff
@@ -5369,6 +5369,17 @@ def subset_fonts(doc: pymupdf.Document, verbose: bool = False, fallback: bool =
...
+
+    Args:
+        fallback: use the older deprecated implementation.
+        verbose: only used by fallback mode.
+
+    Returns:
+        The new MuPDF-based code returns None.  The deprecated fallback
+        mode returns 0 if there are no fonts to subset.  Otherwise, it
+        returns the decrease in fontsize (the difference in fontsize),
+        measured in bytes.
```
cbm755, line 5382: @JorjMcKie please check this "measured in bytes": it based only on reading these two lines:
```python
        new_fontsize += len(new_buffer)

    return old_fontsize - new_fontsize
```

JorjMcKie (approved):

sfadschm:
Should the 'fontTools' requirement then also be removed from the "Optional Features" section of the ReadMe?

jamie-lemon:
> Should the 'fontTools' requirement then also be removed from the "Optional Features" section of the ReadMe?

No as we are still experimenting with the new functionality.

PR #3513, opened 2024-05-23
Addresses wrong label generation

JorjMcKie:
We compute a wrong index for page labels enumerating the ASCII letter range, so that we always deliver label number +1 for a given page number.

This fix subtracts 1 from the index for label styles "a"/"A".

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -4964,7 +4964,9 @@ def get_label_pno(pgNo, labels):
     rule = rule_dict(item)
     prefix = rule.get("prefix", "")
     style = rule.get("style", "")
-    pagenumber = pgNo - rule["startpage"] + rule["firstpagenum"]
+    # make sure we start at 0 when enumerating the alphabet
+    delta = -1 if style in ("a", "A") else 0
+    pagenumber = pgNo - rule["startpage"] + rule["firstpagenum"] + delta
     return construct_label(style, prefix, pagenumber)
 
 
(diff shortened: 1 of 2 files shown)


src/utils.py
```diff
@@ -4964,7 +4964,9 @@ def get_label_pno(pgNo, labels):
     rule = rule_dict(item)
     prefix = rule.get("prefix", "")
     style = rule.get("style", "")
-    pagenumber = pgNo - rule["startpage"] + rule["firstpagenum"]
+    # make sure we start at 0 when enumerating the alphabet
+    delta = -1 if style in ("a", "A") else 0
+    pagenumber = pgNo - rule["startpage"] + rule["firstpagenum"] + delta
```
julian-smith-artifex-com, line 4969: Could we instead change `rule_dict()` to either add one to `rule["startpage"]` or subtract one from `rule["firstpagenum"]`, if it finds a style specification that is 'a' or 'A'? Or would that be doing the wrong thing?

JorjMcKie: I thought about this too, but that would require some other code to be changed. This in the end was the simplest thing to do.
Ultimately I want to replace all this with the native MuPDF page label support, so it simply is not worth investing a lot of thought here.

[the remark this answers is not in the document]
JorjMcKie: As I am writing further below, this is not a piece of code worth to deal a lot with.

Issue #3539, opened 2024-06-01
Add dotted gridline detection to table recognition


JorjMcKie:
### Discussed in https://github.com/pymupdf/PyMuPDF/discussions/3535

<div type='discussions-op-text'>

<sup>Originally posted by **aborruso** May 31, 2024</sup>
Hi,
first of all thank you for this great tool.

I have seen that via python I can use `find_tables`. Is there a way to do something similar via cli and extract a table from a PDF?

Thank you</div>

julian-smith-artifex-com:
Fixed in 1.24.6.

aborruso:
Thank you very much @JorjMcKie and @julian-smith-artifex-com

PR #3540, opened 2024-06-01
Detect dotted gridlines for tables

JorjMcKie:
Addresses #3539

We previously did not detect dotted lines used as table gridlines.

If one of width / height is LE edge_min_length and LT the other dimension, the rectangle is treated as a vertical / horizontal line. We incorrectly used the dimension-specific snap values for this.

We also no longer ignore rectangles if both, width and height are smaller than edge_min_length, but leave this to the snapping and joining algorithms further down the road.


julian-smith-artifex-com (approved):

JorjMcKie:
The reason is that the PDF only has a handful of pages containing images at all: (0-based): 11, 149, 154, 158, 162, 164.
Everything else are vector graphics. They cannot be extracted. All you can do is to find suitable rectangles wrapping them, then make a pixmap of those rectangles and save them as an image file.
We have detection algorithms that are able to identify / cluster atomic graphical commands to larger drawings - see `Page` method [cluster_drawings](https://pymupdf.readthedocs.io/en/latest/page.html#Page.cluster_drawings). The method returns a list of rectangles containing such cluster.
Use this list as described above. There also is a little utility doing this [here](https://github.com/pymupdf/PyMuPDF-Utilities/blob/master/examples/extract-vector-graphics/detect_graphics.py).

As per header / footer exclusion:
This is your job - PDF knows nothing about what these things are. You could do the obvious thing and ignore every bbox returned by `page.cluster_drawings()` which has `y0 <= 72` or `y1 >= page.rect.height - 72`.
Modify 72 as appropriate.

PR #3769, opened 2024-08-11
Correctly locate tessdata from installation folder

JorjMcKie:
If tesseract-ocr is not version 4.00 in a Unix-like platform, we incorrectly used pathlib.Path for locating tessdata.
This fix correctly determines the tessdata folder name for any version of tesseract-ocr.
Note however, that we do not guarantee that MuPDF's own OCR code (which is on Tesseract version 4.00) can cope with tessdata content for a different version than 4.00.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -16,6 +16,7 @@ import io
 import math
 import os
 import pathlib
+import glob
 import re
 import string
 import sys
@@ -18041,7 +18042,7 @@ def get_pdf_str(s: str) -> str:
     return "(" + r + ")"
 
 
-def get_tessdata() -> str:
+def get_tessdata():
     """Detect Tesseract-OCR and return its language support folder.
 
     This function can be used to enable OCR via Tesseract even if the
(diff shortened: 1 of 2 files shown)


src/__init__.py
```diff
@@ -18090,18 +18082,17 @@ def get_tessdata() -> str:
...
-    for sub_response in response.iterdir():
-        for sub_sub in sub_response.iterdir():
-            if str(sub_sub).endswith("tessdata"):
-                tessdata = sub_sub
-                break
-    if tessdata is not None:
+    # search tessdata in folder structure
+    dirname = response[1]  # contains tesseract-ocr installation folder
+    if isinstance(dirname, bytes):  # has probably 'bytes' format
+        dirname = dirname.decode()
```
julian-smith-artifex-com, line 18088: This is ok, but in future we could pass `text=True` to `subprocess.run()` to do this automatically.

src/__init__.py
```diff
@@ -18090,18 +18082,17 @@ def get_tessdata() -> str:
...
+    dirname = pathlib.Path(dirname)  # make a Path from it
+    sub = list(dirname.iterdir())[0]  # includes the version sub-folder
+    tessdata = str(list(sub.iterdir())[0])  # this should be the tessdata
+    if tessdata.endswith("tessdata"):  # expected name suffix
```
julian-smith-artifex-com, line 18092: This is quite difficult to understand, at least for me.

Could we instead do:
```
tessdatas = glob.glob(f'{dirname}/*/tessdata')
if len(tessdatas) == 0:
    message("unexpected: tesseract-ocr has no 'tessdata' folder")
    return False
else:
    # return latest.
    tessdatas.sort()
    return tessdata[-1]
```
?

JorjMcKie: Made both changes.

julian-smith-artifex-com (approved):

PR #3796, opened 2024-08-19
Jules - fixed example code, minor update to docs.

julian-smith-artifex-com:


docs/installation.rst
```diff
@@ -27,6 +27,14 @@ For example:
     python -m pip install --upgrade pip
 
 
+Unsupported platforms
+---------------------------------------------------------
+
+On these platforms, wheels are not provided and PyMuPDF fails to build:
```
jamie-lemon, line 33: Pedantic - PyMuPDF - > |PyMuPDF| , it just means that it will cite our product nicely (see header.rst for code)

docs/recipes-annotations.rst
```diff
@@ -2,6 +2,15 @@
 
 .. _RecipesAnnotations:
 
+.. |toggleStart| raw:: html
```
jamie-lemon, line 5: This is fine, but you haven't wrapped any of the later sample code with `|toggleStart|` & `|toggleEnd|` so it doesn't do anything. Also we may as well move the definition for this into just one place on `header.rst` & remove from `recipes-ocr.rst`, `recipes-multiprocessing.rst`, `recipes-stories.rst` ( I believe those are the only other files which use this component )

jamie-lemon (changes requested):

docs/installation.rst
julian-smith-artifex-com: Ah yes, fixed in my tree.

docs/recipes-annotations.rst
julian-smith-artifex-com: You're quite right about this being unused; i've removed it in my tree.

But we shouldn't put this in `header.rst` or similar, because it is specific to the recipes docs - it contains `<summary><a>See recipe</a></summary>`. I was unable to find a way to specify the text at the place where `toggleStart/End` is used.

jamie-lemon (approved):

PR #3821, opened 2024-08-29
Pixmap.pil() method

vadim0x60:
`Pixmap.pil_save()` and `Pixmap.pil_tobytes()` methods convert a `Pixmap` to a `PIL.Image` object internally, but don't expose this object to the user of the library. Having access to the Pillow object directly is very useful to apply your own image processing to the pixmap or interface with another library that requires Pillow objects as input. I added a method, `Pixmap.pil()` to do exactly that, simply convert a `Pixmap` to a `PIL.Image`


vadim0x60:
recheck

PR #3837, opened 2024-09-03
Jules - extra diagnostics in test_3081

julian-smith-artifex-com:


JorjMcKie (approved):
Linux-specific problem - WOW

PR #3864, opened 2024-09-16
Document v0.0.15 of pymupdf4llm

JorjMcKie:


docs/pymupdf4llm/api.rst
```diff
@@ -28,6 +28,10 @@ The |PyMuPDF4LLM| API
 
     :arg bool write_images: when encountering images or vector graphics, images will be created from the respective page area and stored in the specified folder. Markdown references will be generated pointing to these images. Any text contained in these areas will not be included in the text output (but appear as part of the images). Therefore, if for instance your document has text written on full page images, make sure to set this parameter to `False`.
 
+    :arg bool embed_images: like `write_images`, but images will be included in the markdown text as base64-encoded strings. Ignores `write_images` and `image_path` if used. This may drastically increase the size of your markdown text.
+
+    :arg float image_size_limit: this must be a positive value less than 1. Images are ignored if `width / page.rect.width <= image_size_limit` or `height / page.rect.height <= image_size_limit`.
```
jamie-lemon, line 33: Can we explain more about what the default value of `0.05` means ? Does `1` mean the image is at fun size of the original? Basically need a bit more info about this number!

jamie-lemon (changes requested):

docs/pymupdf4llm/api.rst
JorjMcKie: Ok, I understand. I will update with something like "A value of 0.05 means that to be considered for inclusion, an image's width and height must be larger than 5% of the page's width and height, respectively."

jamie-lemon (approved):

PR #3874, opened 2024-09-18
tests/: test_2608(): (red)add expected result

mjg:
9ccf1c38 ("tests/: test_2608(): expect different output with mupdf master.", 2024-09-17) changed the result for master but meant to leave in the 1.24.x result. Re-add the file from (prior to) that commit at the newly expected location.


julian-smith-artifex-com (approved):
Thanks for this, i was just about to look at it myself but you've beaten me to it.

PR #3878, opened 2024-09-22
Prevent line breaks, deliver reading order.

JorjMcKie:
Refactor plain text and "words" extraction when option sort=True is used:
We previously simply sorted output by ascending bottom and left coordinate. This change collects words  (and respectively text portions) that are approximately on the same line. Apart from extremely malformed pages, words and text are returned in "natural" reading sequence.

This change also suppresses line breaks generated by MuPDF just because of large horizontal distances, as it e.g. often happens between table cell content of the same row.

This change does not alter the user's API and is only activated if `Page.get_text(sort=True)` or `Page.get_text("words", sort=True)` is used.


JorjMcKie:
Have added test data and script.

src/utils.py
```diff
@@ -507,34 +507,195 @@ def get_text_words(
...
+        """
+        line.sort(key=lambda w: w[0].x0)
+        ltext = ""  # text in the line
+        x1 = 0  # end coordinate of ltext
+        lrect = pymupdf.EMPTY_RECT()  # bbox of this line
+        for r, t in line:
+            lrect |= r  # update line bbox
+            # convert distance to previous word to multiple spaces
+            dist = max(
+                int(round((r.x0 - clip.x0 - x1) / r.width * len(t))),
```
julian-smith-artifex-com, line 634: I don't understand why we're using `clip.x0` here. We are interested in distance between words, which is independent of the clip rect, surely?

Also, it looks like we are trying to calculate `<space-between-words> / <average-character-width-in-word>` - is that right? [In which case it might be clearer to write as `round((r.x0 - clip.x0 - x1) / (r.width / len(t))`.

src/utils.py
```diff
@@ -507,34 +507,195 @@ def get_text_words(
...
+    """
+
+    def line_text(clip, line):
+        """Create the string of one text line.
+
+        We are trying to simulate some horizontal layout here, too.
+
+        Args:
+            clip: (pymupdf.Rect) the area from which all text is being read.
+            line: (list) words contained in one text line
```
julian-smith-artifex-com, line 620: I think `line` is a list of (rect, text) tuples?

src/utils.py
```diff
@@ -507,34 +507,195 @@ def get_text_words(
...
+            textpage=textpage,
+            sort=True,
+            tolerance=tolerance,
+        )
+    ]
+
+    if not words:  # no text present
+        return ""
+    totalbox = pymupdf.EMPTY_RECT()  # area covering all text
+    for w in words:
```
julian-smith-artifex-com, line 658: Better to write
```
for rect, text in words:
    totalbox |= rect
```

src/utils.py
```diff
@@ -507,34 +507,195 @@ def get_text_words(
...
+        # if this word matches top or bottom of the line, append it
+        if abs(lrect.y0 - wr.y0) <= tolerance or abs(lrect.y1 - wr.y1) <= tolerance:
+            line.append((wr, text))
+            lrect |= wr
+        else:
+            # output current line and re-initialize
+            ltext = line_text(totalbox, line)
+            lines.append((lrect, ltext))
+            line = [(wr, text)]
+            lrect = wr
```
julian-smith-artifex-com, line 678: This code looks very similar to `sort_words()` above - would it be possible to reuse that code?

src/utils.py
```diff
@@ -507,34 +507,195 @@ def get_text_words(
...
+    # sort all lines vertically
+    lines.sort(key=lambda l: (l[0].y1))
+
+    text = lines[0][1]  # text of first line
+    y1 = lines[0][0].y1  # its bottom coordinate
+    for lrect, ltext in lines[1:]:
+        distance = min(int(round((lrect.y0 - y1) / lrect.height)), 5)
+        breaks = "\n" * (distance + 1)
+        text += breaks + ltext
+        y1 = lrect.y1
```
julian-smith-artifex-com, line 693: [Speculative] This is quite similar to how the new code calculates number of spaces between words, earlier. It might be possible to have a some sort of parameterised function that could be used for horizontal or vertical calculations.

In C++ this would be a template; in Python one would pass small lambda's that return the required info.

This is just me rambling, it has little merit as a review comment.

src/utils.py
JorjMcKie: Checked it: No doesn't make sense.

src/utils.py
JorjMcKie: Without doing a full page layout (as it happens in `pymupdf.exe gettext -mode layout ...`) everything will be approximate only - although this solution here is better than simply putting 1 space between all the words.
The shorter a word (it could be just 1 character), the higher the probability of miscalculating the distance to the predecessor. Everything also depends on font metrics ... in addition to that multiple fonts may be present in one single word ... which is different from MuPDF's notion of a text span.
... And more of that.

src/utils.py
JorjMcKie: We are looking for the distance of the first word from he left border (clip.x0). This causes simulation of line indentation - as opposed to always starting lines in position 0.

But I noticed that the starting value for x1 could be clip.x0, which indeed simplifies the code somewhat.

JorjMcKie:
Now verified that it works for MuPDF v1.24.9 ...

julian-smith-artifex-com (approved):

Issue #3848, opened 2024-09-05
Piximap program crash


JordanGarske:
### Description of the bug

What is happening is that when I read from the PDF, I use the rectangle information to collect color data. Recently, however, I encountered an issue with one of the PDF collections: while collecting the color information, the program crashes unexpectedly without any error messages. I traced the problem to the color_topusage function. Specifically, the error appears to originate from the call to this function, where the issue seems to be related to color_count -> JM_color_count -> oldpix = read_sample(pm, s, n).

### How to reproduce the bug

 I traced the problem to the color_topusage function. Specifically, the error appears to originate from the call to this function, where the issue seems to be related to color_count -> JM_color_count -> oldpix = read_sample(pm, s, n).

### PyMuPDF version

1.24.10

### Operating system

Windows

### Python version

3.12

JorjMcKie:
Please do provide reproducing file and code.

JordanGarske:
[case.pdf](https://github.com/user-attachments/files/16894188/case.pdf)
import pymupdf
from pymupdf import  Page, table, Rect , Pixmap, sRGB_to_rgb

from pymupdf.utils import get_pixmap, get_page_pixmap
#Similar file with error; it loops on a 4-page document and crashes 

on page 2 of 4.
pdf = pymupdf.open(f'./case.pdf')
for i in range(0, len(pdf) ):
    page = pdf.load_page(i)
    print(page)
    for annot in page.get_drawings():
            if page.get_textbox(annot['rect'] ) :
                rect = annot['rect']
                pix: Pixmap = get_pixmap(page, clip= rect)
                colorBytes = pix.color_topusage()
print('done')

JorjMcKie:
This script works:
```python
import pymupdf

doc = pymupdf.open("case.pdf")
page = doc[1]
paths = page.get_drawings()
for i, p in enumerate(paths):
    rect = p["rect"]  # best check for emptiness already here!
    text = page.get_textbox(rect)
    if text:
        pix = page.get_pixmap(clip=rect)
        if pymupdf.IRect(pix.irect).is_empty:
            print(i, "pixmap has empty area - skipping")
            continue
        print("path", i, pix.color_topusage())
```
So the reason for the crash is a pixmap that covers no area, and the `color_topusage()` method does not protext itself against nonsense input.
Which must be fixed of course: we will raise an exception in such cases.

Otherwise please take note of the following comments:
[...]

Issue #3950, opened 2024-10-16
Unable to consistently extract field labels from PDFs


Rutvik-Trivedi:
### Description of the bug

For my usecase, I am trying to extract the `widget.field_label` field from a PDF file. I tried extracting this field from two PDFs. I am successfully able to extract the field labels from one PDF, but not from the other. If it helps in any way, I used [Master PDF Editor](https://code-industry.net/masterpdfeditor/) to add the field labels for the PDFs.

This is the PDF for which I am able to extract the field labels from all the widgets - 
[working sample.pdf](https://github.com/user-attachments/files/17390880/working.sample.pdf)

This is the PDF for which I am not able to extract the field labels even after adding the labels - 
[not working sample.pdf](https://github.com/user-attachments/files/17390896/not.working.sample.pdf)

Is this a PDF/Editor level nuance? Or a bug?

### How to reproduce the bug

The reproduction of the problem should be fairly simple:
```python
import fitz
doc = fitz.Document("working sample.pdf")  # Or "not working sample.pdf"
for page in doc:
    for widget in page.widgets():
        print(widget.field_label)
```
PDF files:
[working sample.pdf](https://github.com/user-attachments/files/17390880/working.sample.pdf)
[not working sample.pdf](https://github.com/user-attachments/files/17390896/not.working.sample.pdf)

For `working sample.pdf`, I get the following output:
```
{{ firstName }}
{{ lastName }}
{{ address.street }}
{{ address.apt }}
{{ address.zipcode }}
{{ address.city }}
{{ spirit }}
{{ today }}
{{ evil | check }}
{{ language.french | X }}
{{ language.esperento | X }}
{{ language.latin | X }}
{{ sig | paste }}
```
Which is correct and expected. It covers all the available field labels

For `not working sample.pdf`, I get the following output:
```
""
None
None
None
```

But the expected output for `not working sample.pdf` should be (not necessarily in the same order):
```
{{ named_insured }}
{{ insurance_line }}
{{ policy_period_start_date }}
{{ policy_period_end_date }}
```
which are all the available field labels in the PDF

### PyMuPDF version

1.24.1

### Operating system

Linux

### Python version

3.10

JorjMcKie:
In this case, the field label is not stored with the field itself, but with its so-called `Parent`. The current code looks at this field `Parent` only for `field_name` while it should also do that for field_label.
The fix is trivial and should be available in a soon new version.

[...]

PR #3979, opened 2024-10-22
src/__init__.py tests/: avoid segv from fz_samples_get() with empty p…

julian-smith-artifex-com:
…ixmap.

Pixmap.color_count(): don't raise exception if JM_color_count() returns empty dict.

_read_samples(): return empty list if pixmap has no samples - avoids segv from fz_samples_get().

Addresses #3848.


julian-smith-artifex-com:
[I've added a commit that addresses #3950.

JorjMcKie (approved):

PR #3997, opened 2024-10-28
Add an example of using Django's file storage API to open files

lancegoyke:
I have been working with pymupdf inside of a Django project and thought it might be useful to document how to use Django's [File Storage API](https://docs.djangoproject.com/en/5.1/ref/files/storage/) to write code using pymupdf that works with a local filesystem and a remote storage backend like S3.

I was able to figure out what to do without these docs based on what is already there, so feel free to close this if you don't think it's necessary.

Also, I have not tested that the docs will build yet. Wasn't sure how to do that and I'm out of time at the moment.

### DIFF
--- a/docs/how-to-open-a-file.rst
+++ b/docs/how-to-open-a-file.rst
@@ -90,6 +90,34 @@ For further examples which deal with files held on typical cloud services please
 ----------
 
 
+Opening Django Files
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Django implements a `File Storage API <https://docs.djangoproject.com/en/5.1/ref/files/storage/>`_ to store files. The default is the `FileSystemStorage <https://docs.djangoproject.com/en/5.1/ref/files/storage/#the-filesystemstorage-class>`_, but the `django-storages <https://django-storages.readthedocs.io/en/latest/index.html>`_ library provides a number of other storage backends.
+
+You can open the file, move the contents into memory, then pass the contents to |PyMuPDF| as a stream.
+
+.. code-block:: python
+
+    import pymupdf
+    from django.core.files.storage import default_storage
+
+    from .models import MyModel
+
+    obj = MyModel.objects.get(id=1)
+    with default_storage.open(obj.file.name) as f:
+        data = f.read()
+
+    doc = pymupdf.Document(stream=data)
+
+Please note that if the file you open is large, you may run out of memory.
+
+The File Storage API works well if you're using different storage backends in different environments. If you're only using the `FileSystemStorage`, you can simply use the `obj.file.name` to open the file directly with |PyMuPDF| as shown in an earlier example.
+
+
+----------
+
+
 
 Opening Files as Text
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~


lancegoyke:

docs/how-to-open-a-file.rst
```diff
@@ -87,6 +87,34 @@ For further examples which deal with files held on typical cloud services please
...
+Django implements a `File Storage API <https://docs.djangoproject.com/en/5.1/ref/files/storage/>`_ to store files. The default is the `FileSystemStorage <https://docs.djangoproject.com/en/5.1/ref/files/storage/#the-filesystemstorage-class>`_, but the `django-storages <https://django-storages.readthedocs.io/en/latest/index.html>`_ library provides a number of other storage backends.
+
+You can open the file, move the contents into memory, then pass the contents to |PyMuPDF| as a stream.
+
+.. code-block:: python
+
+    import pymupdf
+    from django.core.files.storage import default_storage
+
+    from .models import MyModel
```
jamie-lemon, line 105: Where does "MyModel" come from - is that part of Django?

jamie-lemon (changes requested):
Can you add a bit more to this example which explains which Django libraries require to be installed (and how to from pipit I assume)? ( i.e. A basic setup for someone who doesn't know Django would be good ). I would like to be able to test this but don't know Django :)

docs/how-to-open-a-file.rst
lancegoyke: Yes, so `MyModel` is my made up representation of a database table.

lancegoyke:
The set up is definitely non-trivial. It involves installing packages (`django` and `django-storages[s3]`, creating object storage service access keys (I used S3), and adjusting your Django project settings to tell it how to store files. It doesn't work right after installation.

I did create a barebones project if you'd like to test it out for yourself with your own API keys: https://github.com/lancegoyke/pymupdf-django

Perhaps this is more than we care to add in here and the user can fend for her or himself.

The `default_storage` variable is picking whatever configuration the user has already set. If they've gone through the trouble to configure AWS S3 storage for their project, they could be using it as their default storage.

Based on the last hour or two of work on that repo I pasted above, I might recommend we not try to tackle helping the user setup object storage for their project.

Let me know if this is helpful at all or if there's something else I can try to clarify.

jamie-lemon (changes requested):

jamie-lemon:
As I appreciate setting up Django is non-trivial - in this case I think we should just add some further notes about Django and assumed knowledge around it. I know Python developers generally love Dango so I think having this kind of example is valuable!

docs/how-to-open-a-file.rst
```diff
@@ -87,6 +87,34 @@ For further examples which deal with files held on typical cloud services please
 
 
 
+----------
+
+
+Opening Django Files
```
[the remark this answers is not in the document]
lancegoyke: Yes, you are totally correct. I'm embarrassed to see that you saw the original title!

The convention in Django is to call these "media files" or "user-uploaded media". How do you feel about "Opening Media Files in a Django Application"?

Issue #4018, opened 2024-11-04
PyMuPDF hangs when iterating over zero page PDF pages backwards


C-Saunders:
### Description of the bug

PyMuPDF is able to open this PDF, but hangs if I try to iterate over the pages backwards.

### How to reproduce the bug

[zero_pages.pdf](https://github.com/user-attachments/files/17619742/zero_pages.pdf)

```py
import pymupdf

pdf = pymupdf.open("...zero_pages.pdf")

# this raises ValueError: bad start page number
for page in pdf.pages():
    print(page.number)

# this hangs forever
for page in pdf.pages(-1, -1):
    print(page.number)
```

I've tried this on Mac and Linux, and with PyMuPDF versions 1.24.9 and 1.24.13

### PyMuPDF version

1.24.13

### Operating system

MacOS

### Python version

3.11

julian-smith-artifex-com:
Fixed in PyMuPDF-1.24.14.

PR #4019, opened 2024-11-04
Prevent working with empty document

JorjMcKie:
Prevent page iterations if document has zero pages.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -5281,6 +5281,8 @@ class Document:
 
         Arguments have the same meaning as for the range() built-in.
         """
+        if not self.page_count:
+            raise ValueError("document has no pages")
         # set the start value
         start = start or 0
         while start < 0:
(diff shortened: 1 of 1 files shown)


JorjMcKie:
Addresses #4018.

julian-smith-artifex-com (commented):
Not sure we should raise an exception in this case. Could we instead do:
```
if not self.page_count:
    return
```
So that `for page in empty_document.pages(): ...` doesn't raise?

JorjMcKie:
> Not sure we should raise an exception in this case. Could we instead do:
> 
> ```
> if not self.page_count:
>     return
> ```
> 
> So that `for page in empty_document.pages(): ...` doesn't raise?

Sure, why not. It is a very improbable situation anyway: the code lives like this for 5 years or so, without ever this has been happening.

julian-smith-artifex-com (approved):

PR #4035, opened 2024-11-10
unbreak builds against system mupdf

mjg:
Do not determine mupdf version from a local tree if we do not have one. We do not build mupdf in this case anyways.


julian-smith-artifex-com (approved):

julian-smith-artifex-com:
Many thanks for the fix.

PR #4044, opened 2024-11-12
Address 4043

JorjMcKie:
When duplicating a page without /Contents objects, make sure to detect this situation.
Addresses issue 4043.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -4373,7 +4373,7 @@ class Document:
             res = JM_read_contents( page1)
 
             # create new /Contents object for page2
-            if res.m_internal:
+            if res and res.m_internal:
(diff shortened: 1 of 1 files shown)


julian-smith-artifex-com (commented):
JM_read_contents() seems to return either a mupdf buffer if contents is array, or a `bytes`/`str` (from pdf_load_stream()) or empty bytes. So isn't it possible for `if res and res.m_internal:` to still fail if `res` is non-empty str or bytes?

JorjMcKie:
Your opinion?

julian-smith-artifex-com:
Ah yes, i misread pdf_load_stream() as returning a char*. Agree that returning None would be better than b"" but that can be fixed in future.

julian-smith-artifex-com (approved):

PR #4052, opened 2024-11-14
Address bug 4047

JorjMcKie:
Ensure consistent behavior with respect to ignoring meaningless OverlayText.


julian-smith-artifex-com (approved):
[In my tree i've modified _add_redact_annot() to assert that da_str is true if text is true. Which might avoid problems in this area in the future.]

PR #4085, opened 2024-11-25
remove scrubbing of font name

lababidi:
In get_text, I'm expecting the font names to match the same font names from the get_fonts (get_page_fonts). The resulting fonts are different. If the end user wants "scrubbed" font names, applying the scrubbed at the end makes more sense. I don't believe this will break APIs.

For context, having the full font name (AABDFA+Helvetica) is incredibly useful for analyzing PDFs. without it, a lot of the ability is gone.

### DIFF
--- a/.github/workflows/python-package-ubuntu.yml
+++ b/.github/workflows/python-package-ubuntu.yml
@@ -1,0 +1,44 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Python package
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest # ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+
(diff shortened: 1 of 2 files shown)


lababidi:

lababidi:
recheck

lababidi:
@julian-smith-artifex-com any thoughts on this?

JorjMcKie:
I do not think that this PR is able to solve what seems to be your intention.
1. The list `page.get_fonts()` reports the font entries in a page's object definition in `/Resources`. No more - no less. The fonts actually **_used_** may be a true subset of those.
2. The font names appearing in `page.get_text("dict"`) need not (exactly) equal the font names mentioned in that list, and this can never be guaranteed whatever we do. Reasons include:
   * By default, we omit the subset prefix "ABCDEF+" from the font name. You can request to include them by setting a global option `pymupdf.TOOLS.set_subset_fontnames(True)`.
   * The name in "dict" is reported by MuPDF which looks at the self identification in the font file - **_not_** at the PDF object definition. Both names may be (and frequently are) different. Also, MuPDF only returns up to 31 bytes of the font name - the PDF definition may be much longer. So the "dict" output will never be longer than 31 bytes.
   * Text extraction works for all document types - not just PDF. So we cannot expect perfect parallelism with `get_fonts()`, which is a PDF-only function.
   * As a consequence of the above, you may see identical font names in "dict" outputs although the respective text pieces in fact are displayed by truly different fonts in the PDF - even when subset fontnames is `True`.

lababidi:
thank you for the clarification. the global option will help me.

I didn't realize how much integration with mupdf there was with the font system. thanks again.

PR #4110, opened 2024-12-03
Fix a memory issue in `page.get_drawings()`

ZLotusRain:
Fix a memory issue when using `page.get_drawings()` with pdf which contains lots of vector graphics

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -416,7 +416,7 @@ def _format_g(value, *, fmt='%g'):
             return mupdf.fz_format_double(fmt, value)
         else:
             return fmt % value
-        
+
 format_g = _format_g
 
 # Names required by class method typing annotations.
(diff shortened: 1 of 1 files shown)


ZLotusRain:

ZLotusRain:
recheck

JorjMcKie:
Thank you for your submission. I am afraid we are unable to accept your PR however.
* A minor problem is that you are changing many code lines that are not related to the problem you are addressing. While we are aware that the file contains hundreds of lines with trailing spaces, cleaning up code is a different task - and would have to include multiple code quality issues beyond this.
* We did consider in-place updates of that list of dictionaries during development of the method. Our performance tests however have shown that this performs worse than creating a new list. So we decided against this.
* You have not included details concerning your motivation. The additional memory needed by the new list only exists during the creation process and will be released after that. About how many additional MB / GB(?) are we talking here?

PR #4156, opened 2024-12-17
Jules - simplified bool tests, avoid crash in memoryview.

julian-smith-artifex-com:


jonashaag:
Thanks for the quick fix!

I assume it's not possible to keep a reference to the pixmap in the memory view? Because this implementation is still prone to crashes if someone "steals" the `._sample_mv` object.

julian-smith-artifex-com:
Oh, and i think that if someone "steals" the (undocumented, internal) `._sample_mv` object, then they deserve to get a crash...

JorjMcKie (approved):

PR #4216, opened 2025-01-09
Detect image masks in TextPage output

JorjMcKie:
Corresponding documentation updates will follow in a separate PR.


src/extra.i
```diff
@@ -3536,6 +3538,14 @@ void JM_make_image_block(fz_stext_block *block, PyObject *block_dict)
             buf = buffer->buffer;
         } 
         bytes = JM_BinFromBuffer(buf);
+        if (image->mask) {
+            fz_buffer *mask = fz_new_buffer_from_image_as_png(ctx, image->mask, fz_default_color_params);
+            mask_bytes = JM_BinFromBuffer(mask);
+            fz_drop_buffer(ctx, mask);
```
julian-smith-artifex-com, line 3544: We should use `mupdf::fz_new_buffer_from_image_as_png()` to get a FzBuffer instead of a fz_buffer*, so avoiding the explicit call to `fz_drop_buffer`, and avoiding a leak if JM_BinFromBuffer() raises an exception. [JM_BinFromBuffer() is overloaded to take a fz_buffer* or a FzBuffer.]

src/__init__.py
```diff
@@ -12600,19 +12605,20 @@ def extractIMGINFO(self, hashes=0):
...
+            block_dict[dictkey_colorspace] = mupdf.fz_colorspace_n(cs)
+            block_dict[dictkey_cs_name] = mupdf.fz_colorspace_name(cs)
+            block_dict[dictkey_xres] = img.xres()
+            block_dict[dictkey_yres] = img.yres()
+            block_dict[dictkey_bpc] = img.bpc()
+            block_dict[dictkey_size] = img_size
             if hashes:
-                block_dict[ "digest"] = digest
+                block_dict["digest"] = digest
+            block_dict["has-mask"] = has_mask
```
julian-smith-artifex-com, line 12621: Isn't `has-mask` redundant - can't the user simply use `block_dict['mask'] is not None`?

JorjMcKie: I could have used "mask" as key name. Didn't do it to avoid the misconception that this **_is_** the mask. This method in general does not return image binary data at all - only metadata.
So this key name represents an effort to make sure it is information **_about_** the image - not the image itself.

JorjMcKie: but your reaction in itself show that it is a confusing choice. But we definitely should use bool values.

PR #4222, opened 2025-01-13
Document mask in TextPage

JorjMcKie:
This is a document only PR to show how transparent images are identified in TextPage output


jamie-lemon (approved):
It will affect the corresponding Japanese version of the these pages as translation strings will not find a match in many areas, so I'll need to follow up with an update to fix this. Please remind me if I don't do this by the end of the week!

PR #4237, opened 2025-01-18
Support miter

JorjMcKie:


JorjMcKie:
When writing text using `Page.insert_text()` or `Page.insert_textbox()` using render mode != 0, text appearance may show "spikes" when unfortunate combinations of font size and border-width parameters are chosen.
This fix introduces a new parameter `miter_limit` with a default value that suppresses all such spikes. Its value can be changed to either other (> 1) numbers or to `None` to allow PDF defaults.

julian-smith-artifex-com (approved):

PR #4244, opened 2025-01-20
Jules - Address signed/unsigned bug in handling of span color and added alpha value.

julian-smith-artifex-com:


src/extra.i
```diff
@@ -3052,11 +3051,11 @@ mupdf::FzRect JM_make_spanlist(
         FZ_STEXT_STROKED = 32,
         FZ_STEXT_CLIPPED = 64
         */
-        int char_flags;
+        unsigned char_flags;
         #endif
         
         const char *font = "";
-        unsigned int color = -1;
+        unsigned int argb = -1;
```
JorjMcKie, line 3058: A negative value for unsigned does not make much sense, does it? I think we can just use 0.

JorjMcKie (approved):

Issue #4254, opened 2025-01-26
add_freetext_annot is drawing text outside the annotation box


apalermo01:
### Description of the bug

I originally put this issue is sioyek [here](https://github.com/ahrm/sioyek-python-extensions/issues/26) but I think this is an issue with pymupdf itself. 

Annotations generated using `add_freetext_annot` sometimes appear invisible until one of its properties is changed in a pdf viewer.

### How to reproduce the bug

I set up this script that reproduces the issue:

```python
import fitz

pdf_path = "/home/alex/Documents/tutorial.pdf"
output_path =  "/home/alex/Documents/output.pdf"

doc = fitz.open(pdf_path)
page = doc[0]

rect = fitz.Rect(100, 100, 200, 150)
annot = page.add_freetext_annot(rect, "Test Annotation from minimal example")
annot.set_border(width=1, dashes=(3,3))
annot.set_opacity(0.5)
annot.set_colors(stroke=(1, 0, 0), fill=(0.9, 0.9, 0.9))
# annot.update()

rect = fitz.Rect(200, 200, 300, 300)
annot2 = page.add_freetext_annot(rect, "Test Annotation from minimal example pt 2")
annot2.set_border(width=1, dashes=(3,3))
annot2.set_opacity(0.5)
annot2.set_colors(stroke=(1, 0, 0), fill=(0.9, 0.9, 0.9))
annot2.update()

print("annot 1 AP:")
print(annot._getAP())
print("annot 2 AP:")
print(annot2._getAP())

doc.save(output_path)
```

`annot` appears on the page when I open the file in okular, but `annot2` appears invisible. When I go in okular, select the annotation, and change any of its properties, it appears.  I attached a video demonstrating this:

https://github.com/user-attachments/assets/1d3f2d8a-3c92-487a-92b3-d80703897d54

Here are the appearance strings that we printed when I ran this:

```bash
annot 1 AP:
[...]

PR #4265, opened 2025-02-03
Address 4254

JorjMcKie:
For bug #4254, ensure that multiple FreeText annotations are displayed as expected.

We also add the following new features:
1. Support for subtype FreeTextCallout
2. Support for rich text.


jamie-lemon (approved):
Docs LGTM.

PR #4272, opened 2025-02-06
open memoryview object as-is they were bytes

arthurdarcet:
it supports the same protocol and avoids an unecessary copy


arthurdarcet:

arthurdarcet:
recheck

JorjMcKie:
I don't think that this change makes sense.
The base library does ultimately require a memory area to work with, it cannot use `memoryview` objects directly.
So we would have to convert a memoryview object to `bytes` before calling MuPDF. This conversion (`bytes(mv)`, respectively `mv.tobytes()`) however **_does_** make a copy and thus negates any assumed advantage.
I think that for a memoryview object `mv` passing `bytes(mv)` to the constructor is a negligible effort.

arthurdarcet:
Could you point me where that copy happens in my fork? The change I
proposed works as is, without any extra copy as far as I can see.
I'm also not sure how the mv would behave any differently from a bytes
object, apart from what type(obj) returns of, hence my proposed change, as
that's more or less how the python doc define mem views

Le sam. 8 févr. 2025, 14:36, Jorj X. McKie ***@***.***> a
écrit :

JorjMcKie:
@arthurdarcet - sorry, you are right:
I wasn't aware of that the C-code had already been changed to directly support Python's buffer protocol. Therefore all binary formats - `bytes`, `memoryview` and `bytearray` - are readily supported without the need for extra conversions.
Could you please use the `isinstance()` function to perform the checks - instead of using `type(stream)`?

arthurdarcet:
@JorjMcKie of course - just pushed that. I also moved some tests a bit, but logic should still be exactly the same.
I left the assert on line 2933 - but i don't think it's needed? It doesn't hurt, but it will always pass. Let me know if you want me to remove it

PR #4285, opened 2025-02-11
update pyodide-build version

linuxrider:
This minimal change should allow building for latest pyodide.
The culprit was the export option `pyinit`. Changing it to `whole_archive` solves it for me.
I do not know what the implications of this change are.
https://pyodide.org/en/stable/development/meta-yaml.html#build-exports
Thanks to @ryanking13 for the hint in pyodide/pyodide#5111.

### DIFF
--- a/scripts/test.py
+++ b/scripts/test.py
@@ -474,7 +474,7 @@ def build_pyodide_wheel(pyodide_build_version=None):
     #
     env_extra['PYMUPDF_SETUP_MUPDF_TESSERACT'] = '0'
     setup = pyodide_setup(pymupdf_dir, pyodide_build_version=pyodide_build_version)
-    command = f'{setup} && pyodide build --exports pyinit'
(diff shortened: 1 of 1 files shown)


linuxrider:

linuxrider:
recheck

ryanking13:
_Copying my response at https://github.com/pyodide/pyodide/issues/5111._

Changing to `--export whole_archive` means that the library will expose all the symbols to the public, which will increase the output size and memory usage a bit. I think it won't be a very big deal for most of the libraries, but FYI.

By default (--export pyinit), the pyodide will only export symbols that start with PyInit, which Python C-Extension libraries would need to expose. But in the PyMuPDF case, it was building the shared library libmupdf.so as well in the build script. Since libmupdf.so itself is not a C-Extension Python library, nothing was exposed, hence the empty output.

julian-smith-artifex-com (approved):

PR #4374, opened 2025-03-14
adjust to python 3.14

mjg:
`typing.ByteString` has been deprecated since python 3.12 will be removed in python 3.14. For now, use a union type alias rather than identifying the minimal interface.

Fedora being early birds again - but what a day for that!

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -419,8 +419,15 @@ def _format_g(value, *, fmt='%g'):
         
 format_g = _format_g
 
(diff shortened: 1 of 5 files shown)


cbm755:
FYI, linking to downstream issue https://bugzilla.redhat.com/show_bug.cgi?id=2352547

happy :pie: day!

julian-smith-artifex-com (approved):
Thanks for this, i've checked locally that it works Python-3.14.

PR #4491, opened 2025-05-07
Jules - improved document constructor, check mupdf file opening, removed support for mupdf-1.24.

julian-smith-artifex-com:


src/__init__.py
JorjMcKie, line 1: Could we include all text extraction flag bits?
image

JorjMcKie (changes requested):

JorjMcKie (approved):

PR #4493, opened 2025-05-07
Docs cleanup

jamie-lemon:


JorjMcKie (approved):

jamie-lemon:
I'll be putting in the Japanese equivalents following these syntax updates next. The swill take a few days as it is pretty arduous ....
After that it should be ready.

PR #4500, opened 2025-05-14
Table Detection Improvements

JorjMcKie:
* Support new detection parameter "add_boxes" which allows specifying "virtual" rectangles to help detection.

* Support new parameter "paths" to allow specifying previously extracted vector graphics.

* Several minor improvements, especially we now export line breaks inside table cells as HTML "<br>" tags instead of replacing "\n" by spaces.

### DIFF
--- a/src/table.py
+++ b/src/table.py
@@ -79,6 +79,7 @@ import html
 from collections.abc import Sequence
 from dataclasses import dataclass
 from operator import itemgetter
+import weakref
 
 # -------------------------------------------------------------------
 # Start of PyMuPDF interface code
@@ -87,6 +88,8 @@ from . import (
     Rect,
     Matrix,
     TEXTFLAGS_TEXT,
+    TEXT_FONT_BOLD,
+    TEXT_FONT_SUPERSCRIPT,
     TOOLS,
     EMPTY_RECT,
     sRGB_to_pdf,
@@ -1061,7 +1064,7 @@ def words_to_edges_v(words, word_threshold: int = DEFAULT_MIN_WORDS_VERTICAL):
         if not overlap:
             condensed_bboxes.append(bbox)
 
-    if len(condensed_bboxes) == 0:
+    if not condensed_bboxes:
         return []
 
     condensed_rects = map(bbox_to_rect, condensed_bboxes)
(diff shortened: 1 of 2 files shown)


src/table.py
```diff
@@ -1462,22 +1487,33 @@ def _get_header(self, y_tolerance=3):
...
             """
-            for b in page.get_text("dict", flags=TEXTFLAGS_TEXT, clip=bbox)["blocks"]:
-                for l in b["lines"]:
-                    for s in l["spans"]:
-                        if s["flags"] & 16:
-                            return True
-            return False
+            blocks = page.get_text("dict", flags=TEXTFLAGS_TEXT, clip=bbox)["blocks"]
+            spans = [s for b in blocks for l in b["lines"] for s in l["spans"]]
+            return any([bool(s["flags"] & 16) for s in spans])  # pylint: disable=R1729
```
julian-smith-artifex-com, line 1516: Can we use `TEXT_FONT_BOLD` here instead of `16`?

src/table.py
```diff
@@ -1489,44 +1525,57 @@ def top_row_is_bold(bbox):
...
+        # 1-column tables have no extra header
         if len(cells) < 2:
             return header_top_row
 
-        col_x = [
-            c[2] if c is not None else None for c in cells[:-1]
-        ]  # column (x) coordinates
+        # assume top row is the header if second row is empty
+        row2 = self.rows[1]  # second row
+        if all([c is None for c in row2.cells]):  # pylint: disable=R1729
```
julian-smith-artifex-com, line 1538: Similar to above, suggest generator expression.

src/table.py
```diff
@@ -1489,44 +1525,57 @@ def top_row_is_bold(bbox):
...
-                        spans.append(s)
+        blocks = page.get_text("dict", clip=clip, flags=TEXTFLAGS_TEXT)["blocks"]
+        # non-empty, non-superscript spans above table, sorted descending by y1
+        spans = sorted(
+            [
+                s
+                for b in blocks
+                for l in b["lines"]
+                for s in l["spans"]
+                if not (white_spaces.issuperset(s["text"]) or s["flags"] & 1)
```
julian-smith-artifex-com, line 1569: Change `s["flags"] & 1` to `s["flags"] & TEXT_FONT_ITALIC` ?

src/table.py
JorjMcKie: Actually, the suggestion R1729 does not make much sense, because making a generator is unnecessary overhead for these small numbers of checks.
But I will replace this with a normal loop from which an early return exit is possible. More code of course, but what don't we do for the satisfaction of pylint?
The symbolic names for the hardcoded numbers are certainly reasonable.

src/table.py
JorjMcKie: I will use `if set(row2.cells) == {None}:` which does what is needed here.

src/table.py
JorjMcKie: Sure: symbolic names instead of hardcoded numbers. The meaning of this is "SUPERSCRIPT" BTW.

src/table.py
julian-smith-artifex-com: I can't see the problem with using generator expressions.

* According to https://peps.python.org/pep-0289/, compared to list comprehensions, performance with generator expressions is "roughly comparable for small to mid-sized data sets". And they are faster as the data set gets larger.

* The code using generator expressions is slightly smaller than the code that uses list comprehensions, and avoids the pylint warning.

* The code using generator expressions is much simpler than using an explicit loop.

* I don't know whether the explicit loop might be faster or not. I suspect any difference (whether faster or slower) would not be significant for PyMuPDF.

However if you really want to avoid generator expressions, that's ok by me too!

src/table.py
julian-smith-artifex-com: Really? That seems both obscure and inefficient.

julian-smith-artifex-com: Hmm, that seems a bit obscure, but i'm ok with it if you insist.

julian-smith-artifex-com: Interesting, the opposite of what i expected!

JorjMcKie: Admittedly, this only plays any role for very long sequences. If however the list is really long and there is a high chance for non-equality in the first items, the generator reacts with shortcut exit and thus is preferrable.
All this under the assumption, that the list / sequence in question is not being created as part of the generator expression.

Just tried with setting item 100 to something else (not `None`). Then the generator was slightly faster than the set building.

JorjMcKie: I think there are a lot of pros and cons. Let me make another change and use generators wherever possible. This will avoid doubts and discussions about the best way each time again.

Forgot to mention another advantage of generators: they do not depend on sequence items being hashable (in contrast to sets).

julian-smith-artifex-com (approved):

PR #4510, opened 2025-05-19
Address 4505

JorjMcKie:
Add an option for Widget updates that will propagate the field flags to all kids under the same parent.


src/__init__.py
```diff
@@ -7173,6 +7173,51 @@ def _validate(self):
...
+        its /Kids widgets to the value of the current widget.
+        Only possible for widgets existing in the PDF.
+
+        Returns True or False.
+        """
+        if not self.xref:
+            return False  # no xref: widget not in the PDF
+        doc = self.parent.parent  # the owning document
+        if not doc:  # should never happen
+            return False
```
julian-smith-artifex-com, line 7189: Could we do `assert(doc)` here?

src/__init__.py
```diff
@@ -7173,6 +7173,51 @@ def _validate(self):
...
+        if not Parent.pdf_is_dict():
+            return False  # no /Parent: nothing to do
+
+        # put the field flags value into the parent field flags:
+        Parent.pdf_dict_put_int(PDF_NAME("Ff"), self.field_flags)
+
+        # also put that value into all kids of the Parent
+        kids = Parent.pdf_dict_get(PDF_NAME("Kids"))
+        if not kids.pdf_is_array():
+            return False  # no /Kids: should never happen!
```
julian-smith-artifex-com, line 7203: Could we use an assert here? Our return value is not checked by the caller, so this will silently fail.

JorjMcKie: I am reluctant here:
After all, this may be an ill-formed PDF. We might want to issue a warning instead.
The other case (doc not available) would be a genuine PyMuPDF error.

julian-smith-artifex-com (approved):

PR #4516, opened 2025-05-22
Update table.py

BBC-Esq:
Parameters
        ----------
        footnote : {"none", "last_single_cell"}, optional
            If "last_single_cell" and the table’s final row has exactly one
            non-empty cell, that row is treated as a foot-note.

        Returns
        -------
        list  |  (list, str | None)


BBC-Esq:

PR #4517, opened 2025-05-22
Update table.py

BBC-Esq:
Adds header_margin and footer_margin keyword-arguments to page.find_tables(), letting callers ignore running headers/footers without manually building a clip rectangle. Parameters after page are now keyword-only to keep all existing code that omits the new args fully backward-compatible.


BBC-Esq:

PR #4523, opened 2025-05-24
Address 4520

JorjMcKie:
We previously refused a source page without a /Contents object in method `show_pdf_page`.
This fix creates a Form XObject in the target PDF with a stream length of 0 in this case.

### DIFF
--- a/tests/test_4520.py
+++ b/tests/test_4520.py
@@ -1,0 +1,19 @@
+import pymupdf
+
+
+def test_4520():
+    """Accept source pages without /Contents object in show_pdf_page."""
+    vsn_tuple = tuple(map(int, pymupdf.__version__.split(".")))
+    tar = pymupdf.open()
+    src = pymupdf.open()
+    src.new_page()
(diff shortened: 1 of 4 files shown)


tests/test_4520.py
```diff
@@ -0,0 +1,19 @@
+import pymupdf
+
+
+def test_4520():
+    """Accept source pages without /Contents object in show_pdf_page."""
+    vsn_tuple = tuple(map(int, pymupdf.__version__.split(".")))
```
julian-smith-artifex-com, line 6: You could use `pymupdf.pymupdf_version_tuple` here.

However we don't generally try to make tests work with different versions of PyMuPDF. For example there will be test failures if one does `pip install pymupdf==1.25.5; pytest PyMuPDF`.

It's different with mupdf versions where we often look at `pymupdf.mupdf_version_tuple` so we can test the current PyMuPDF with different versions of MuPDF (generally the current release branch (1.26.x) and master).

Issue #4527, opened 2025-05-26
Rect.intersects() is much slower than necessary


stijnvermeeren-swisstopo:
### Description of the bug

When benchmarking one of my scripts, I noticed that the method `Rect.intersects()` (which is called thousands of times) is a significant bottleneck.
I was able to replace the PyMuPDF method with a simple custom implementation which behaves the same but is more than 10x faster.

Is there a good reason for the current, slower PyMuPDF implementation, or could this method be made much faster?

Example output of the reproduction script below:
```
$ time python test.py 
PyMuPDF time: 7.175560788000526
Fast method time: 0.551379568999675
```

### How to reproduce the bug

```
import pymupdf
import random
import time

def random_rect() -> pymupdf.Rect:
    x0 = random.randint(0, 100)
    y0 = random.randint(0, 100)
    width = random.randint(0, 100)
    height = random.randint(0, 100)
    return pymupdf.Rect(x0, y0, x0 + width, y0 + height)

def fast_intersects(rect1: pymupdf.Rect, rect2: pymupdf.Rect) -> bool:
    return not rect1.is_empty and not rect2.is_empty and (
        (rect1.x0 < rect2.x1) and (rect2.x0 < rect1.x1) and (rect1.y0 < rect2.y1) and (rect2.y0 < rect1.y1)
    )

pymupdf_time = 0
fast_time = 0

random_rects = [random_rect() for _ in range(1000)]
for rect1 in random_rects:
    for rect2 in random_rects:
        a = time.process_time()
        pymupdf_result = rect1.intersects(rect2)
        b = time.process_time()
        fast_result = fast_intersects(rect1, rect2)
        c = time.process_time()

        pymupdf_time += b - a
        fast_time += c - b
        assert(fast_result == pymupdf_result)

print(f"PyMuPDF time: {pymupdf_time}")
print(f"Fast method time: {fast_time}")
```

### PyMuPDF version

1.26.0

### Operating system

Linux

### Python version

3.12

julian-smith-artifex-com:
Fixed in PyMuPDF-1.26.3.

Issue #4462, opened 2025-04-22
delete_pages() does not accept a single int


JojoMueller:
### Description of the bug

document.delete_pages() should accept a single int positional parameter - called "Format 3" in documentation. But calling it with a single int parameter triggers an exception when trying to covert it to  a tuple with tuple(r). 

### How to reproduce the bug

call e.g. delete_pages(-1)

### PyMuPDF version

1.25.4

### Operating system

Windows

### Python version

3.13

JorjMcKie:
Thanks for your report. Fix developed.

julian-smith-artifex-com:
Fixed in PyMuPDF-1.26.3.

JojoMueller:
Just installed and tested the fix. The command to reproduce the bug `delete_pages(-1)`, which is a valid "Format 3" statement, still doesn't work.  See: 
` if numbers[0] < 0 or numbers[-1] >= page_count:
            raise ValueError("bad page number(s)")`

JorjMcKie:
> Just installed and tested the fix. The command to reproduce the bug `delete_pages(-1)`, which is a valid "Format 3" statement, still doesn't work. See: ` if numbers[0] < 0 or numbers[-1] >= page_count: raise ValueError("bad page number(s)")`

Thanks for the observation.~~We will prohibit negative numbers via documentation.~~

julian-smith-artifex-com:
Fixed in PyMuPDF-1.26.4.

PR #4561, opened 2025-06-13
Minor Changes

JorjMcKie:
Speed up rectangle intersection check (issue #4527): avoid calculating an intersection rectangle.

Address #4462: Support single int as argument.


JorjMcKie:
The changed `intersects()` algorithm is on average 3.8 times faster, mainly because it no longer builds the actual intersection rectangle to check its emptiness.

PR #4563, opened 2025-06-16
Documentation: Elaborates on Table related classes.

jamie-lemon:


docs/page.rst
```diff
@@ -2329,56 +2329,169 @@ This is an overview of homologous methods on the :ref:`Document` and on the :ref
...
 
-   An object representing a table found on the page. Attributes of interest:
+   An object representing a table found on the page.
+
+
+   .. attribute:: page
+
+      A description of the page instance for the table.
+
+      :type: `string`
```
JorjMcKie, line 2386: :type: :ref:`Page`

docs/page.rst
```diff
@@ -2329,56 +2329,169 @@ This is an overview of homologous methods on the :ref:`Document` and on the :ref
...
 
 .. class:: TableHeader
 
-.. class:: TableRow
 
+   Dedicated class for table headers.
 
+   .. attribute:: bbox
 
+      The bounding box of the table given as a tuple `(x0, y0, x1, y1)`. This is the rectangle that contains all cells of the table.
```
JorjMcKie, line 2465: The bounding box of the union of cells belonging to the table header, given as a tuple `(x0, y0, x1, y1)`. This rectangle contains all table header cells.

JorjMcKie (changes requested):
There are several corrections ...

docs/page.rst
JorjMcKie: This `page` attribute is technically speaking a "weakref proxy" of the `Page` object. Python has this construct to allow such "weak" back-references. They behave like the original object (methods, attributes), but will react gracefully (with an exception - not with a Python interpreter crash) if the object has died in the meantime.
Our approach in this way takes care of situations where users might extract and collect tables across multiple pages: this is forbidden! The only way to achieve this kind of thing is collecting the extracted outputs from `to_markdown`, `to_pandas` etc.

JorjMcKie (approved):

PR #4615, opened 2025-07-16
Intercept erroneous widgets

JorjMcKie:
When copying widgets as part of insert_pdf() processing, badly defined source widgets cause MuPDF exceptions when building their grafted object version. This fix intercepts MuPDF exceptions and ignores such widgets.

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -1987,7 +1987,11 @@ def do_widgets(
                 w_obj.pdf_dict_get(pymupdf.PDF_NAME("Parent"))
             )
             if parent_xref == 0:  # parent not in target yet
-                w_obj_graft = mupdf.pdf_graft_mapped_object(graftmap, w_obj)
+                try:
+                    w_obj_graft = mupdf.pdf_graft_mapped_object(graftmap, w_obj)
+                except Exception as e:
+                    pymupdf.message_warning(f"cannot copy widget at {xref=}: {e}")
+                    continue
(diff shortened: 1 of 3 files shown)


src/utils.py
```diff
@@ -1987,7 +1987,11 @@ def get_acroform(doc):
                 w_obj.pdf_dict_get(pymupdf.PDF_NAME("Parent"))
             )
             if parent_xref == 0:  # parent not in target yet
-                w_obj_graft = mupdf.pdf_graft_mapped_object(graftmap, w_obj)
+                try:
+                    w_obj_graft = mupdf.pdf_graft_mapped_object(graftmap, w_obj)
+                except Exception as e:
```
julian-smith-artifex-com, line 1992: Is there a specific exception type that we could catch here so that we don't hide unexpected errors?

tests/test_4614.py
```diff
@@ -0,0 +1,11 @@
...
+import os
+
+
+def test_4614():
+    script_dir = os.path.dirname(__file__)
+    filename = os.path.join(script_dir, "resources", "test_4614.pdf")
+    src = pymupdf.open(filename)
+    doc = pymupdf.open()
+    doc.insert_pdf(src)
+    wtext = pymupdf.TOOLS.mupdf_warnings(doc)
```
julian-smith-artifex-com, line 11: We should assert that `wtext` contains the expected warnings.

julian-smith-artifex-com (approved):

PR #4642, opened 2025-08-05
tests/test_general.py: fix expected svg output in test_3569 from MuPD…

sebras:
…F 1.27.x.


sebras:
> Does this test now fail with the 1.26.x branch?

d'oh! yes, of course it will. that's not a good patch!

sebras:
Ok, that's better. I could separate out the prefix and suffix strings that are identical, but that code would be more complicated and susceptible to mistakes. These version dependent tests usually just duplicate the strings with the necessary differences, so that's what I opted for.

julian-smith-artifex-com (approved):

PR #4647, opened 2025-08-11
docs: fix conf after _build.py changes

mjg:
`docs/conf.py` tries to read version info from `src/__init__py` using an expression match. This breaks after e439a0fa ("setup.py src/ tests/: auto-generate pymupdf.pymupdf_version and add git info.", 2025-08-08) makes `src/__init__py` import that info from `_build.py`, which is generated during build in the build directory.

Make the build process generate a copy of `_build.py` in `docs` and let `docs/conf.py` import version info from there.

Alternative fixes:
- pointing `docs/conf.py` at the build directory; this may no longer exist after a wheel build, though
- reading from the generated wheel; oh well ;-)
- installing the build and importing `_build.py` from there; this would mean intermingling build and install steps

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 build/
 demo/README.rst
 docs/build
+docs/_build.py
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -49,16 +49,8 @@ copyright = "2015-" + str(thisday.year) + ", Artifex"
 # built documents.
 #
 # The full version, including alpha/beta/rc tags.
-_path = os.path.abspath(f'{__file__}/../../src/__init__.py')
-with open(_path) as f:
-    for line in f:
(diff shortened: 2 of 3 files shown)


julian-smith-artifex-com:
Thanks for this. Oddly enough i came across the same problem earlier today.

My fix is slightly different - it makes `docs/conf.py` import PyMuPDF's `setup.py` and then use `setup.version_p` directly.

I think this is slightly better because it doesn't require that a build has been done. And it gets the required information directly from the primary source `setup.py`.

mjg:
If that has no unwanted side-effects it should be fine. I saw no commit fixing this, that's why I went ahead.

mjg:
Works with your patch, btw, so I'll close this one. Thanks!

julian-smith-artifex-com:
Thanks, we've merged the docs fix.

PR #4736, opened 2025-10-10
Insert a space between two separated non-space characters.

ChangHyukSoo:
Two non-space characters were separated in the PDF file, but they were not separated when I get texts using page.get_text().
By modifying JM_print_stext_page_as_text() in src/extra.i , I resolved this case.
I attach a sample PDF file for you to reproduce my case.
[2023_insert_18.pdf](https://github.com/user-attachments/files/22841368/2023_insert_18.pdf)
Example codes:
  import pymupdf
  doc = pymupdf.open('2023_insert_18.pdf')
  page = doc[0]
  text = page.get_text()
  print(text)

Before modification: "2023년중세계경제는고금리･고물가지속,"
After modification:  "2023년 중 세계경제는 고금리･고물가 지속,"   <-- This is the correct result.

### DIFF
--- a/src/extra.i
+++ b/src/extra.i
@@ -2308,6 +2308,8 @@ void JM_print_stext_page_as_text(mupdf::FzBuffer& res, mupdf::FzStextPage& page)
             for (auto line: block)
             {
                 int last_char = 0;
+                int prev_c = 0;         // 2025.10.10
+                float prev_x = 0.0f;    // 2025.10.10
                 for (auto ch: line)
                 {
                     fz_rect chbbox = JM_char_bbox( line, ch);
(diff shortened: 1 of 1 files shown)


JorjMcKie:
Thank you again for your thoughts and effort!
Text extraction is implemented in our base library MuPDF. Detecting whether and when to insert a space between adjacent characters is part of a carefully chosen threshold there.
If we want or need to change anything here it must happen in the base library not in PyMuPDF. There also are a lot of things to consider which you did not cover in your suggestion - for example the different output variants beyond plain text. After all there are "words", "blocks", "dict", "html" etc. which need to be covered one way or another ... and which could dramatically suffer (e.g. what was one word previously could now become 2 different words).

I've just talked to the MuPDF team. Their response was clear: global changes to the separator threshold value must be handled with extreme care because of the high risk breaking stuff elsewhere.
Potentially we might provide a parameter that eases adjustments in borderline cases.

You can always make your own logic and compose text based on single character extractions, i.e. "rawdict". That's all we have in our hands, sorry.

I hope for your understanding.

PR #4748, opened 2025-10-15
Support pymupdf_layout

JorjMcKie:
Completely new version

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -28,7 +28,7 @@ import weakref
 import zipfile
 
 from . import extra
-
+import importlib.util
 
 # Set up g_out_log and g_out_message from environment variables.
 #
@@ -332,6 +332,37 @@ class _Globals:
         self.skip_quad_corrections = 0
 
 _globals = _Globals()
+
+_get_layout: typing.Optional[typing.Callable] = None
+
+# global switch ensuring that the recommendation message is shown at most once
+_recommend_layout = True  # must be referred to as "global" everywhere
+
+
+def no_recommend_layout():
+    """For users who never want to see the layout recommendation."""
+    global _recommend_layout
+    _recommend_layout = False
+
+
+def _warn_layout_once():
+    """Check if we should recommend installing the layout package."""
+    msg="""Consider using the pymupdf_layout package for a greatly improved page layout analysis."""
+
+    global _recommend_layout
+    if (
+        1
+        and _recommend_layout  # still True?
+        and _get_layout is None  # no layout function stored here
+
+        # client did not globally disable the recommendation
+        and os.getenv("PYMUPDF_SUGGEST_LAYOUT_ANALYZER") != "0"
+
+        # layout is not available in this Python
+        and not importlib.util.find_spec("pymupdf.layout")
+    ):
+        print(msg)
+        _recommend_layout = False  # never show the message again
 
 
 # Optionally use MuPDF via cppyy bindings; experimental and not tested recently
@@ -9952,7 +9983,7 @@ class Page:
         return rc
 
     def _get_textpage(self, clip=None, flags=0, matrix=None):
-        if g_use_extra:
+        if 1 or g_use_extra:
             ll_tpage = extra.page_get_textpage(self.this, clip, flags, matrix)
             tpage = mupdf.FzStextPage(ll_tpage)
             return tpage
@@ -10781,6 +10812,20 @@ class Page:
         pclip = JM_rect_from_py(clip)
         mupdf.pdf_clip_page(pdfpage, pclip)
 
+    def get_layout(self):
+        """Try to access layout information."""
+
+        if self.layout_information is not None:
+            # layout information already present
+            return
+
+        if not _get_layout:
+            # no layout information available
+            return
+
+        layout_info = _get_layout(self)
+        self.layout_information = layout_info
+
     @property
     def artbox(self):
         """The ArtBox"""
@@ -11432,7 +11477,7 @@ class Page:
         assert isinstance(page, mupdf.FzPage), f'{self.this=}'
         clips = True if extended else False
         prect = mupdf.fz_bound_page(page)
-        if g_use_extra:
+        if 1 or g_use_extra:
             rc = extra.get_cdrawings(page, extended, callback, method)
         else:
             rc = list()
@@ -12157,7 +12202,7 @@ class Page:
             self.set_rotation(0)
         page = self.this
         rc = []
-        if g_use_extra:
+        if 1 or g_use_extra:
             dev = extra.JM_new_texttrace_device(rc)
         else:
             dev = JM_new_texttrace_device(rc)
@@ -13205,6 +13250,9 @@ class Page:
         return self.parent.page_xref(self.number)
 
     rect = property(bound, doc="page rectangle")
+
+    # any result of layout analysis is stored here
+    layout_information = None
 
 
 class Pixmap:
@@ -16391,7 +16439,7 @@ class TextPage:
 
     def extractBLOCKS(self):
         """Return a list with text block information."""
-        if g_use_extra:
+        if 1 or g_use_extra:
             return extra.extractBLOCKS(self.this)
         block_n = -1
         this_tpage = self.this
@@ -16587,7 +16635,7 @@ class TextPage:
 
     def extractWORDS(self, delimiters=None):
         """Return a list with text word information."""
-        if g_use_extra:
+        if 1 or g_use_extra:
             return extra.extractWORDS(self.this, delimiters)
         buflen = 0
         last_char_rtl = 0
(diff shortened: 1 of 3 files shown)


src/table.py
```diff
@@ -80,11 +81,12 @@
 from dataclasses import dataclass
 from operator import itemgetter
 import weakref
+import pymupdf
 
 # -------------------------------------------------------------------
 # Start of PyMuPDF interface code
 # -------------------------------------------------------------------
-from . import (
+from pymupdf import (
```
julian-smith-artifex-com: It probably doesn't matter, but why this change? Did `from . import (...)` stop working for some reason?

src/table.py
```diff
@@ -2543,6 +2639,25 @@ def find_tables(
         "text_x_tolerance": text_x_tolerance,
         "text_y_tolerance": text_y_tolerance,
     }
+
+    old_quad_corrections = pymupdf.TOOLS.unset_quad_corrections()
```
julian-smith-artifex-com: We should do:
```
old_quad_corrections = pymupdf.TOOLS.unset_quad_corrections()
try:
    ...
finally:
    pymupdf.TOOLS.unset_quad_corrections(old_quad_corrections)
```
Otherwise global state gets modified silently if an exception is thrown.

src/extra.i
```diff
@@ -3638,38 +3799,132 @@ static void JM_make_text_block(fz_stext_block *block, PyObject *block_dict, int
...
+
+void _getTableBlocks(fz_stext_page *tp, PyObject *table_dict, PyObject *bbox, int dev_flags)
 {
     fz_context* ctx = mupdf::internal_context_get();
+    fz_rect bounds = JM_rect_from_py(bbox);
     fz_stext_block *block;
     fz_buffer *text_buffer = fz_new_buffer(ctx, 128);
-    PyObject *block_dict, *block_list = PyList_New(0);
+    PyObject *block_list = PyList_New(0);
+    block = fz_find_table_within_bounds(ctx, tp, bounds);
```
julian-smith-artifex-com: This looks very unsafe if fz_find_table_within_bounds() raises a mupdf exception - it'll longjmp into nowhere.

At the very least we should call `mupdf::ll_fz_find_table_within_bounds()` so that mupdf exceptions are converted to c++ exceptions. But even then it'll leak resources.

julian-smith-artifex-com (changes requested):
As mentioned elsewhere, i'm worried that this PR adds a large amount of C code, some of which is unsafe if mupdf or C++ exceptions are thrown.

If we are sure that all this C code is required for speed reasons, then that's ok, but we need to fix the various resource cleanup issues i think.

src/table.py
```diff
@@ -93,26 +95,116 @@
...
 CHARS = []  # text characters from PyMuPDF
 TEXTPAGE = None
 TEXT_BOLD = mupdf.FZ_STEXT_BOLD
 TEXT_STRIKEOUT = mupdf.FZ_STEXT_STRIKEOUT
-FLAGS = TEXTFLAGS_TEXT | TEXT_COLLECT_STYLES
-
+FLAGS = TEXTFLAGS_TEXT | TEXT_COLLECT_STYLES | TEXT_ACCURATE_BBOXES | TEXT_MEDIABOX_CLIP
+FLAGS2 = (
+    0 | TEXT_ACCURATE_BBOXES | TEXT_SEGMENT | TEXT_COLLECT_VECTORS | TEXT_MEDIABOX_CLIP
+)
```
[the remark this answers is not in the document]
JorjMcKie: It doesn't break the tests obviously.
Otherwise, because of additional tables that are now detected by layout, we are dependent on accurate bboxes to ensure correct functioning of MuPDF's table structure detector under all possible situations of documents in the wild.

src/extra.i
```diff
@@ -3603,7 +3689,81 @@ void JM_make_image_block(fz_stext_block *block, PyObject *block_dict)
...
+        block = fz_find_table_within_bounds(ctx, tp, bounds);
+
+        // Check if a table structure was found
+        if (block && block->type == FZ_STEXT_BLOCK_GRID)
+        {
+            JM_make_grid_block(block, table_dict);
+        }
+    }
+    fz_catch(ctx)
+        {;}
```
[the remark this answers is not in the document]
julian-smith-artifex-com: In that case for clarity could we do this?:
```
    try
    {
            block = mupdf:ll_fz_find_table_within_bounds(tp, bounds);
    }
    catch (std::exception&)
    {
        /* Ignore failure to find a table structure. */
        return;
    }
    // Check if a table structure was found
    if (block && block->type == FZ_STEXT_BLOCK_GRID)
    {
        JM_make_grid_block(block, table_dict);
    }
```

src/__init__.py
```diff
@@ -13206,6 +13243,9 @@ def xref(self):
 
     rect = property(bound, doc="page rectangle")
 
+    # any result of layout analysis is stored here
+    layout_information: typing.Optional[typing.List[tuple]] = None
```
julian-smith-artifex-com: Do we actually need the typing hint  here? It makes this simple line of code look very complicated, and presumably this is internal implementation, not for users?

src/__init__.py
```diff
@@ -333,6 +333,29 @@ def __init__(self):
...
+    global _recommend_layout
+    if (
+        1
+        and _recommend_layout
+        and not callable(_get_layout)
+        and os.getenv("PYMUPDF_SUGGEST_LAYOUT_ANALYZER") != "0"
+        and not importlib.util.find_spec("pymupdf.layout")
+    ):
+        print(msg)
+        _recommend_layout = False
```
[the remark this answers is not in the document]
JorjMcKie: Yes, all these statements `global _recommend_layout` are needed.

src/__init__.py
```diff
@@ -10781,6 +10804,20 @@ def clip_to_rect(self, rect):
         pclip = JM_rect_from_py(clip)
         mupdf.pdf_clip_page(pdfpage, pclip)
 
+    def get_layout(self, vertical_gap=12):
```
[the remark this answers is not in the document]
JorjMcKie: Please leave in there - may be required later.

src/__init__.py
JorjMcKie: Ok ... then let's simplify this ...

src/extra.i
```diff
@@ -3320,35 +3370,28 @@ PyObject* extractWORDS(mupdf::FzStextPage& this_tpage, PyObject *delimiters)
                                 );
                     }
                     mupdf::fz_clear_buffer(buff);
-                    buflen = 0;  // reset char counter
-                    if (word_delimiter) continue;
+                    buflen = 0;
                 }
```
julian-smith-artifex-com: Could we use the size of `buff`'s data here (e.g. from `fz_buffer_storage()`) instead of maintaining this separate measure? I guess there could be a small overhead, but it would be simpler, clearer and less error-prone.

tests/test_textextract.py
```diff
@@ -787,7 +787,8 @@ def test_extendable_textpage():
         # Create a pymupdf.TextPage() for <stext_page> so we can use
         # text_page.extractDICT() etc.
         text_page = pymupdf.TextPage(stext_page)
-        
+        text_page._dev_flags = stext_options.flags
+
```
julian-smith-artifex-com: This seems to be the only occurrence of `_dev_flags`, can it be removed?

src/table.py
```diff
@@ -80,39 +80,120 @@
...
-    TEXT_FONT_BOLD,
-    TEXT_FONT_ITALIC,
-    TEXT_FONT_MONOSPACED,
-    TEXT_FONT_SUPERSCRIPT,
-    TEXT_COLLECT_STYLES,
-    TOOLS,
-    EMPTY_RECT,
-    sRGB_to_pdf,
-    Point,
-    message,
```
julian-smith-artifex-com: I like this change :-)

src/__init__.py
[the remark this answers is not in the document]
julian-smith-artifex-com: In that case could we add a comment?

tests/test_linequad.py
```diff
@@ -11,12 +12,14 @@
...
 def test_quadcalc():
+    old_settings = pymupdf.TOOLS.unset_quad_corrections()
+    pymupdf.TOOLS.unset_quad_corrections(False)
     text = " angle 327"  # search for this text
     doc = pymupdf.open(filename)
     page = doc[0]
     # This special page has one block with one line, and
     # its last span contains the searched text.
-    block = page.get_text("dict", flags=0)["blocks"][0]
+    block = page.get_text("dict")["blocks"][0]
```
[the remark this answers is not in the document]
JorjMcKie: Was an unnecessary change: everything works both ways. I'll take back the change.

julian-smith-artifex-com:
Thanks for the various updates.

One last thing: the patch has three trivial whitespace changes to files in the tests/ directory. I think it would be worthwhile removing these changes, then the patch can be trivially seen to not make any changes to our test suite.

julian-smith-artifex-com (approved):

Issue #4756, opened 2025-10-19
swig --version doesn't work in all versions of swig; -version should be used instead


ngie-eign:
Did code inspection and it looks like `setup.py` relies on `swig --version` returning something meaningful (see the linked code below).

`--version` isn't implemented on all versions of swig (it seems that it's only technically implemented on the MacOS version of the package).

### FreeBSD
```
% swig --version
Unable to find option or file '--version', Use 'swig -help' for more information.
% swig -version 

SWIG Version 4.2.1

Compiled with c++ [amd64-portbld-freebsd14.3]

Configured options: +pcre

Please see https://www.swig.org for reporting bugs and further information
```

### Linux
```
% lsb_release -a
No LSB modules are available.
Distributor ID: Raspbian
Description:    Raspbian GNU/Linux 12 (bookworm)
Release:        12
Codename:       bookworm
% swig --version
Unable to find option or file '--version', Use 'swig -help' for more information.
% swig -version 

SWIG Version 4.1.0

Compiled with g++ [arm-unknown-linux-gnueabihf]

Configured options: +pcre

Please see https://www.swig.org for reporting bugs and further information
```

### macOS
```
% swig --version

SWIG Version 4.3.1

Compiled with clang++ [aarch64-apple-darwin24.2.0]

Configured options: +pcre

Please see https://www.swig.org for reporting bugs and further information
```

https://github.com/pymupdf/PyMuPDF/blob/5b6acfb369fb4c460ffeb5f2288d6efe81b93c93/setup.py#L729

Arthur-Barreto:
We’re running into the same deprecation warnings on Python 3.12:

DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute
DeprecationWarning: builtin type SwigPyObject has no __module__ attribute
DeprecationWarning: builtin type swigvarlink has no __module__ attribute

They show up as soon as we import fitz, both in our unit tests and in a container image we ship. Right now we’re silencing them via filterwarnings, but just wanted to +1 that the problem affects us too. Happy to retest when a fix lands.

julian-smith-artifex-com:
We keep issues open until they are fixed in a release, so reopening.

ngie-eign:
> We keep issues open until they are fixed in a release, so reopening.

Sounds good - thanks!

julian-smith-artifex-com:
Fixed in pymupdf-1.26.6.

PR #4770, opened 2025-10-29
setup.py: use swig -version instead of swig --version

ngie-eign:
`swig --version` doesn't seem to be supported in all versions of swig, e.g., FreeBSD and some flavors of Linux. Use `swig -version` instead when querying the CLI tool.

Closes: #4756


ngie-eign:

ngie-eign:
recheck

julian-smith-artifex-com:
Thanks for this.

We've not actually had problems from `swig --version` with pypi.org's swig on Windows, Linux and MacOS. However the swig docs only mention `-version` and it's good to fix problems on other platforms anyway.

PR #4785, opened 2025-11-12
tests/test_general.py: test_3569(): update svg font name to match mup…

julian-smith-artifex-com:
…df master.


julian-smith-artifex-com:
Closing because mupdf is being updated to avoid the change.

PR #4786, opened 2025-11-12
Use uuid, it speeds up the call

yuval-peleg-levy-ai21:
Use uuid, it speeds up the call


JorjMcKie:
I've reviewed your proposal, but I don't believe it addresses a significant performance concern.

Our current implementation achieves durations in the low single-digit microsecond range — even with name lists of around 20 entries, which is itself uncommon. When using a set() for containment checks, performance nearly doubles and matches or exceeds that of a UUID-based approach when the list size is 10 or fewer — the most typical case.

Additionally, the current naming strategy produces short and readable identifiers in contrast to UUIDs like 'b58cb87d-50a0-4ae2-be88-77ea77afdb0d'.

Given these factors, we’ll stick with the existing solution and ensure we will use the `set()` of names to avoid duplication.

yuval-peleg-levy-ai21:
Hey,
The loop isn’t the main problem, but the getting of pdf info. I have a use case where I’m inserting a lot of htmls, and it sped it up by a factor of 3. 

I did this after seeing these calls are a bottlenck using a profiler. 

Thanks!

JorjMcKie:
Ah, now I understand your motivation better!
Hm, we will reconsider this within our team.

yuval-peleg-levy-ai21:
Thanks. Sorry for not being more detailed in my PR.

PR #4812, opened 2025-11-26
Updates workflow .yaml to try to ignore draft PRs.

jamie-lemon:
Changes made:
Added `if: github.event.pull_request.draft == false || github.event_name == 'workflow_dispatch'` to both jobs. This condition:

- Skips the job if the PR is a draft
- Still allows the workflow to run when manually triggered via workflow_dispatch

The condition is necessary for both triggers because workflow_dispatch doesn't have a pull_request object, so we need the `|| github.event_name == 'workflow_dispatch'` part to avoid errors if manually running the workflow.

### DIFF
--- a/.github/workflows/test_quick.yml
+++ b/.github/workflows/test_quick.yml
@@ -15,6 +15,7 @@ jobs:
 
   master:
     name: mupdf master
+    if: github.event.pull_request.draft == false || github.event_name == 'workflow_dispatch'
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


julian-smith-artifex-com (commented):
I'd prefer to leave this file unchanged.

A simpler solution would be to make a PR target a dummy thread, not main. In this case test_quick will not be run.

julian-smith-artifex-com:
Closing this as agreed elsewhere.

PR #4819, opened 2025-11-29
Update helper-select.i

Mohataseem89:
1. added missing Py_DECREF for early-continue branch
2. PyLong_FromLong(i) instead of cast
3. added necessary comments 4.improved readability


JorjMcKie:
Sorry, we no longer maintain the classic version of PyMuPDF and are therefore unable to accept Pull Requests for it.

PR #4855, opened 2026-01-05
Update PyMuPDF to cope with warnings changes in 1.28.0.

robinwatts:
Text extraction now only produces warnings about corrupt struct trees or layers, if the struct trees or layers are used.


robinwatts:
The failure with master and the PR is expected, and will be solved as soon as this goes in.

PR #4909, opened 2026-02-17
Add __iter__ method to Document class for proper iteration support

matheushrc:
## Description

This PR adds the `__iter__()` method to the `Document` class to make it properly iterable and satisfy type checkers (like Pylance/Pyright) when using the class with `enumerate()` and other iteration patterns.

## Problem

While the `Document` class already implements `__getitem__()` and `__len__()`, which technically makes it a sequence, type checkers require an explicit `__iter__()` method to recognize it as `Iterable[Page]`. This causes type checking errors when using patterns like:

```python
for page_num, page in enumerate(doc):
    # Type error: Argument of type 'Document' cannot be assigned to parameter 'iterable' of type 'Iterable[_T@enumerate]'
```

## Solution

Added a simple `__iter__()` method that yields pages by leveraging the existing `__getitem__()` and `__len__()` methods:

```python
def __iter__(self):
    """Iterate over all pages in the document."""
    for i in range(len(self)):
        yield self[i]
```

## Testing

- The existing test suite already uses `enumerate(doc)` and `for page in doc:` patterns extensively (20+ instances found across test files)
- These tests will automatically verify the `__iter__()` implementation
- No breaking changes - this only adds functionality that was implicitly expected

## Benefits

- Resolves type checking errors when iterating over Document objects
- Makes the Document class explicitly iterable per Python protocols
- Maintains backward compatibility
- Efficient implementation using existing methods


matheushrc:

julian-smith-artifex-com:
Isn't this a bug in Pylance/Pyright?

As you point out, `pymupdf.Document` is already a sequence.

And from https://docs.python.org/3/library/functions.html#enumerate:

> enumerate(iterable, start=0)
> ... iterable must be a sequence, an iterator, or some other object which supports iteration. ...

So type checkers should not be matching against only `Iterable[Page]` here, surely?

PR #4955, opened 2026-03-26
Fix incorrect lineJoin and pathfactor in get_drawings()

### LINKED ISSUE #4954: get_drawings() returns incorrect lineJoin and width
### Description of the bug

`get_drawings()` returns incorrect `lineJoin` and `width` values due to two bugs in `jm_lineart_stroke_path`:

1. **`lineJoin` is multiplied by `pathfactor`**: `stroke->linejoin` is an enum (0=Miter, 1=Round, 2=Bevel) and should not be scaled. Note that `lineCap` is correctly handled as plain integers without `pathfactor` multiplication, so this is inconsistent.

2. **`pathfactor` only handles uniform scaling and 90° rotation**: For non-uniform scaling (e.g. `2 0 0 3 0 0 cm`), `pathfactor` falls back to 1, making `width` incorrect.

This bug is hard to notice because `linejoin=0` (Miter, the most common value) always produces `0 * pathfactor = 0`.

Affected code:
- `src/extra.i`: `jm_lineart_stroke_path`
- `src_classic/helper-devices.i`: same function

### Bug 1: lineJoin is scaled by pathfactor

```python
import fitz

doc = fitz.open()
page = doc.new_page(width=200, height=200)
shape = page.new_shape()
shape.draw_line((0, 0), (1, 1))
shape.finish(color=(0, 0, 0), width=0.1)
shape.commit()
content = b"q\n0.12 0 0 0.12 0 0 cm\n2 j\n6 w\n100 100 m\n800 100 l\nS\nQ\n"
doc.update_stream(page.get_contents()[0], content)
pdf_bytes = doc.tobytes()
doc.close()

doc2 = fitz.open(stream=pdf_bytes, filetype="pdf")
d = doc2[0].get_drawings()[-1]
print(f"lineJoin={d['lineJoin']}")  # Expected: 2, Actual: 0.24
doc2.close()
```

| CTM scale | linejoin (raw) | Expected | Actual |
|-----------|---------------|----------|--------|
| 0.12 | 2 (Bevel) | 2 | 0.24 |
| 2.0 | 2 (Bevel) | 2 | 4.0 |
| 2.83 | 1 (Round) | 1 | 2.83 |

For comparison, `lineCap` is correctly returned as integers without scaling:
```c
// lineCap - correct (no pathfactor)
Py_BuildValue("iii", stroke->start_cap, stroke->dash_cap, stroke->end_cap)

// lineJoin - incorrect (pathfactor applied)
Py_BuildValue("f", dev->pathfactor * stroke->linejoin)
```

### Bug 2: width is wrong with non-uniform scale

```python
doc = fitz.open()
page = doc.new_page(width=200, height=200)
shape = page.new_shape()
shape.draw_line((0, 0), (1, 1))
shape.finish(color=(0, 0, 0), width=0.1)
shape.commit()
content = b"q\n2 0 0 3 0 0 cm\n1 w\n10 10 m\n90 10 l\nS\nQ\n"
doc.update_stream(page.get_contents()[0], content)
pdf_bytes = doc.tobytes()
doc.close()

doc2 = fitz.open(stream=pdf_bytes, filetype="pdf")
d = doc2[0].get_drawings()[-1]
print(f"width={d['width']}")  # Expected: 2.0, Actual: 1.0
doc2.close()
```

### Suggested fix

**lineJoin** — use the raw enum value without scaling:
```c
// Before (src/extra.i)
DICT_SETITEMSTR_DROP(dev->pathdict, "lineJoin",
    Py_BuildValue("f", dev->pathfactor * stroke->linejoin));

// After
DICT_SETITEMSTR_DROP(dev->pathdict, "lineJoin",
    Py_BuildValue("i", stroke->linejoin));
```

**pathfactor** — use `sqrt(a² + b²)` to handle arbitrary transforms:
```c
// Before (src/extra.i)
dev->pathfactor = 1;
if (ctm.a != 0 && fz_abs(ctm.a) == fz_abs(ctm.d))
    dev->pathfactor = fz_abs(ctm.a);
else if (ctm.b != 0 && fz_abs(ctm.b) == fz_abs(ctm.c))
    dev->pathfactor = fz_abs(ctm.b);

// After
float scale = sqrtf(ctm.a * ctm.a + ctm.b * ctm.b);
if (scale < 1e-9f)
    scale = sqrtf(ctm.c * ctm.c + ctm.d * ctm.d);
if (scale < 1e-9f)
    scale = 1.0f;
dev->pathfactor = scale;
```

Both `src/extra.i` and `src_classic/helper-devices.i` have the same bugs.

Note: For non-uniform scaling, stroke width is direction-dependent in general.
This fix approximates it using the length of the transformed unit vector.

### PyMuPDF version

1.27.2.2

### Operating system

Linux

### Python version

3.13

sakamotch:
Fixes #4954

### Bug 1: `lineJoin` scaled by `pathfactor`

`stroke->linejoin` is an enum (0=Miter, 1=Round, 2=Bevel) but was multiplied by `pathfactor`. Changed to emit the raw integer value, consistent with how `lineCap` is already handled.

```c
// Before
Py_BuildValue("f", dev->pathfactor * stroke->linejoin)

// After
Py_BuildValue("i", stroke->linejoin)
```

### Bug 2: `pathfactor` falls back to 1 for non-uniform scaling

The original calculation only covered uniform scaling (`|a|==|d|`) and 90° rotation (`|b|==|c|`). Changed to use `sqrt(a² + b²)` which handles arbitrary transforms.

```c
// Before
dev->pathfactor = 1;
if (ctm.a != 0 && fz_abs(ctm.a) == fz_abs(ctm.d))
    dev->pathfactor = fz_abs(ctm.a);
else if (ctm.b != 0 && fz_abs(ctm.b) == fz_abs(ctm.c))
    dev->pathfactor = fz_abs(ctm.b);

// After
float scale = sqrtf(ctm.a * ctm.a + ctm.b * ctm.b);
if (scale < 1e-9f)
    scale = sqrtf(ctm.c * ctm.c + ctm.d * ctm.d);
if (scale < 1e-9f)
    scale = 1.0f;
dev->pathfactor = scale;
```

Note: For non-uniform scaling, stroke width is direction-dependent in general. This fix approximates it using the length of the transformed unit vector.

### Files changed

- `src/extra.i` — rebased build
- `src_classic/helper-devices.i` — classic build
- `src/__init__.py` — Python fallback

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -23208,11 +23208,7 @@ def jm_lineart_stroke_path( dev, ctx, path, stroke, ctm, colorspace, color, alph
     #log(f'{dev.pathdict=} {dev.clips=}')
     try:
         assert isinstance( ctm, mupdf.fz_matrix)
-        dev.pathfactor = 1
-        if ctm.a != 0 and abs(ctm.a) == abs(ctm.d):
-            dev.pathfactor = abs(ctm.a)
-        elif ctm.b != 0 and abs(ctm.b) == abs(ctm.c):
-            dev.pathfactor = abs(ctm.b)
+        dev.pathfactor = math.sqrt(abs(ctm.a * ctm.d - ctm.b * ctm.c))
         dev.ctm = mupdf.FzMatrix( ctm)  # fz_concat(ctm, dev_ptm);
         dev.path_type = trace_device_STROKE_PATH
 
@@ -23228,7 +23224,7 @@ def jm_lineart_stroke_path( dev, ctx, path, stroke, ctm, colorspace, color, alph
                 stroke.dash_cap,
                 stroke.end_cap,
                 )
-        dev.pathdict[ 'lineJoin'] = dev.pathfactor * stroke.linejoin
+        dev.pathdict[ 'lineJoin'] = float(stroke.linejoin)
         if 'closePath' not in dev.pathdict:
             #log('setting dev.pathdict["closePath"] to false')
             dev.pathdict['closePath'] = False
--- a/src/extra.i
+++ b/src/extra.i
@@ -2719,14 +2719,7 @@ jm_lineart_stroke_path(fz_context *ctx, fz_device *dev_, const fz_path *path,
     jm_lineart_device *dev = (jm_lineart_device *)dev_;
     //printf("extra.jm_lineart_stroke_path(): dev->seqno=%zi\n", dev->seqno);
     int i;
-    dev->pathfactor = 1;
-    if (ctm.a != 0 && fz_abs(ctm.a) == fz_abs(ctm.d)) {
-        dev->pathfactor = fz_abs(ctm.a);
-    } else {
-        if (ctm.b != 0 && fz_abs(ctm.b) == fz_abs(ctm.c)) {
-            dev->pathfactor = fz_abs(ctm.b);
-        }
-    }
+    dev->pathfactor = sqrtf(fabsf(ctm.a * ctm.d - ctm.b * ctm.c));
     dev->ctm = ctm; // fz_concat(ctm, trace_device_ptm);
     dev->path_type = STROKE_PATH;
 
@@ -2739,7 +2732,7 @@ jm_lineart_stroke_path(fz_context *ctx, fz_device *dev_, const fz_path *path,
     DICT_SETITEMSTR_DROP(dev->pathdict, "color", jm_lineart_color(colorspace, color));
     DICT_SETITEM_DROP(dev->pathdict, dictkey_width, Py_BuildValue("f", dev->pathfactor * stroke->linewidth));
     DICT_SETITEMSTR_DROP(dev->pathdict, "lineCap", Py_BuildValue("iii", stroke->start_cap, stroke->dash_cap, stroke->end_cap));
-    DICT_SETITEMSTR_DROP(dev->pathdict, "lineJoin", Py_BuildValue("f", dev->pathfactor * stroke->linejoin));
+    DICT_SETITEMSTR_DROP(dev->pathdict, "lineJoin", Py_BuildValue("f", (float)stroke->linejoin));
     if (!PyDict_GetItemString(dev->pathdict, "closePath")) {
         DICT_SETITEMSTR_DROP(dev->pathdict, "closePath", JM_BOOL(0));
     }


sakamotch:

src/__init__.py
```diff
@@ -23194,11 +23194,12 @@ def jm_lineart_stroke_path( dev, ctx, path, stroke, ctm, colorspace, color, alph
...
-        if ctm.a != 0 and abs(ctm.a) == abs(ctm.d):
-            dev.pathfactor = abs(ctm.a)
-        elif ctm.b != 0 and abs(ctm.b) == abs(ctm.c):
-            dev.pathfactor = abs(ctm.b)
+        scale = math.sqrt(ctm.a ** 2 + ctm.b ** 2)
+        if scale < 1e-9:
+            scale = math.sqrt(ctm.c ** 2 + ctm.d ** 2)
+        if scale < 1e-9:
+            scale = 1.0
+        dev.pathfactor = scale
```
julian-smith-artifex-com: Thanks for creating this PR. I have a couple of questions:

How come we convert very small scales (<1e-9) to 1 instead of zero or 1e-9? Do we divide by the scale elsewhere?

It seems a little strange to replaces one partial expression (which uses .a and .d, or .b and .c, or none of them) with a different partial expression (which uses .a and .b, or .c and .d, or none of them).

Could we instead start with the determinant as a scalar representing the matrix? The determinant is how a matrix scales areas, which suggests we use `sqrt(abs(determinant))` as the matrix scale, i.e. `math.sqrt(abs(ctm.a*ctm.d-ctm.b*ctm.c))`.

src_classic/helper-devices.i
```diff
@@ -449,7 +453,7 @@ jm_lineart_stroke_path(fz_context *ctx, fz_device *dev_, const fz_path *path,
 	DICT_SETITEMSTR_DROP(dev_pathdict, "color", jm_lineart_color(ctx, colorspace, color));
 	DICT_SETITEM_DROP(dev_pathdict, dictkey_width, Py_BuildValue("f", dev_pathfactor * stroke->linewidth));
 	DICT_SETITEMSTR_DROP(dev_pathdict, "lineCap", Py_BuildValue("iii", stroke->start_cap, stroke->dash_cap, stroke->end_cap));
-	DICT_SETITEMSTR_DROP(dev_pathdict, "lineJoin", Py_BuildValue("f", dev_pathfactor * stroke->linejoin));
+	DICT_SETITEMSTR_DROP(dev_pathdict, "lineJoin", Py_BuildValue("f", (float)stroke->linejoin));
 	if (!PyDict_GetItemString(dev_pathdict, "closePath")) {
 		DICT_SETITEMSTR_DROP(dev_pathdict, "closePath", JM_BOOL(0));
 	}
```
julian-smith-artifex-com, line 455: `src_classic/` is no longer used and is only retained in git as a reference (we should probably remove it) and probably does not build these days.

So it's probably best to remove this part of the patch.

src/__init__.py
sakamotch: Thanks for the suggestion! I'm now testing the determinant-based approach with my PDFs. Will follow up with the results.

src_classic/helper-devices.i
sakamotch: Makes sense, removed!

src/__init__.py
sakamotch: Verified with my PDFs, everything looks good!

The determinant-based approach correctly handles uniform scaling with arbitrary rotation angles, not just 0° and 90° like the original code:

For a uniform scale `s` with any rotation angle `θ` (where `a = s·cosθ, b = s·sinθ, c = -s·sinθ, d = s·cosθ`):
- `sqrt(|a·d - b·c|) = sqrt(s²·cos²θ + s²·sin²θ) = s` ✓

For non-uniform scaling, it gives the geometric mean of the two scale factors, which is a reasonable approximation.

For degenerate matrices (zero determinant), pathfactor becomes 0, meaning width=0. Since strokes are invisible in this case, no fallback to 1 is needed.

Does this look good to you?

julian-smith-artifex-com:
Many thanks for updating things.

Could your squash your 4 commits into 1 commit and rebase on top of current main?

Then i think i'd be ok to merge this PR, after checking with @JorjMcKie.

sakamotch:
Thanks for the review.
Squashed into one commit and rebased onto current `main`.

JorjMcKie (approved):

julian-smith-artifex-com (approved):

sakamotch:
I checked the CI failure more carefully. The determinant-based `pathfactor` I introduced here is the problem.

I'll change it to:

```c
float sx = sqrtf(ctm.a * ctm.a + ctm.b * ctm.b);
float sy = sqrtf(ctm.c * ctm.c + ctm.d * ctm.d);
float dot = ctm.a * ctm.c + ctm.b * ctm.d;

if (sx == sy && dot == 0.0f)
    dev->pathfactor = sx;
else
    dev->pathfactor = 1.0f;
```

This extends the old logic from 0/90-degree cases to arbitrary rotation angles, while keeping the same strict fallback behavior.

I’m keeping `==` intentionally, since the previous implementation already used exact float comparisons.

I tested this locally and `tests/test_annots.py::test_redact4` now passes, and `tests/test_drawings.py` also passes with this change.

sakamotch:
@julian-smith-artifex-com If this revised approach looks good, I can update the PR with a new squashed commit.

julian-smith-artifex-com:
test_redact4 is currently failing due to a regression in mupdf in the latest master branch. So i don't think you should try to get the test to pass here.

Tests are passing with this PR and git branch 1.27.x, so i think the PR is good.

julian-smith-artifex-com:
[rerunning tests with mupdf master as failure was from mupdf, not thie PR.]

julian-smith-artifex-com:
I hope you don't mind, but i've taken the liberty of rebasing the PR tree on top of latest main.

Hopefully it will still pass tests, and we can merge the PR.

sakamotch:
Thanks again for taking care of the earlier rebase.

It looks like `main` moved forward again, so I rebased the PR branch onto the latest `main`. Hope that's okay, and hopefully it should be mergeable now.

sakamotch:
@julian-smith-artifex-com Checks passed, looks mergeable now!

PR #4975, opened 2026-04-23
Jules - various changes for next release.

julian-smith-artifex-com:


docs/conf.py
```diff
@@ -52,23 +52,43 @@
...
-    del sys.path[0]
-python_versions_minor = test.python_versions_minor
-del test
+if 1:
+    # Importing setup.py requires pipcl etc so instead of importing, we grep
+    # for the version info in setup.py and scripts/test.py.
+    setup_py_path = os.path.normpath(f'{__file__}/../../setup.py')
+    with open(setup_py_path) as f:
+        setup_py_text = f.read()
+    regex = "\nversion_p = '([0-9.]+)'\n"
```
JorjMcKie, line 61: Isn't a r in front of the string missing? Like ` regex = r"\nversion_p = '([0-9.]+)'\n"`

JorjMcKie (commented):
I think there are some syntax issues around back-slashed strings for regex having to "r" prefix.

docs/conf.py
julian-smith-artifex-com: Ah yes, you're quite right, thanks. Actually there are some test failures so i've converted the PR to draft and running more tests.

JorjMcKie: Ah sorry, wrong trip

julian-smith-artifex-com: [Actually i tested the doc generation a while ago, and it worked; have just tried again and it still works, so i think this is ok.]

JorjMcKie (approved):

PR #4984, opened 2026-05-04
Japanese and Korean docs no longer build on RTD due to Python version

jamie-lemon:
See information from RTD:

Screenshot 2026-05-04 at 22 10 20

I believe this relates to the Python version defined in requirements.txt

### DIFF
--- a/docs/locales/ja/.readthedocs.yaml
+++ b/docs/locales/ja/.readthedocs.yaml
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: "3.9"
+    python: "3.13"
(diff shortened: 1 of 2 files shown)


jamie-lemon:
Build failure info from RTD

[ja-RTD-build-failure.txt](https://github.com/user-attachments/files/27376925/ja-RTD-build-failure.txt)

JorjMcKie (approved):

julian-smith-artifex-com (commented):
Suggest we just use `python:` so that we never have to worry about this again.

jamie-lemon:
Let's see if this fixes it then I'll follow up with one which does that for all our .yamls

PR #5015, opened 2026-06-06
remove usage of typing.ByteString

mjg:
typing.ByteString is deprecated in python 3.15 and slated for removal in python 3.17. This creates deprecation warnings which trip up our tests.

Since we are requiring python >= 3.10 already, we can switch to the proper alternative unconditionally now.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -462,12 +462,7 @@ def _format_g(value, *, fmt='%g'):
         
 format_g = _format_g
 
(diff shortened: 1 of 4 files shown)


mjg:
This has been applied (partially at least) with different commit message in c7a2cf16

julian-smith-artifex-com:
Apologies for not noticing this PR. The fix was made after adding testing with python-3.15.

PR #5026, opened 2026-06-17
Docs: Updates with an example on how to check for black & white pages.

jamie-lemon:
Also some updates to file conversion section & docs README

### DIFF
--- a/docs/the-basics.rst
+++ b/docs/the-basics.rst
@@ -1085,9 +1085,58 @@ Another example could be redacting an area of a page, but not to redact any line
 Converting PDF Documents
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-We recommend the pdf2docx_ library which uses |PyMuPDF| and the **python-docx** library to provide simple document conversion from |PDF| to **DOCX** format.
+See :doc:`converting-files` for more information.
 
 
+.. note::
+
+    **PDF -> DOCX**
+    
+    We recommend the pdf2docx_ library which uses |PyMuPDF| and the **python-docx** library to provide simple document conversion from |PDF| to **DOCX** format.
+
+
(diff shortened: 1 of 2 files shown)


docs/the-basics.rst
```diff
@@ -1085,7 +1085,58 @@ Another example could be redacting an area of a page, but not to redact any line
...
+        results = []
+        for i, page in enumerate(doc):
+            pix=page.get_pixmap(colorspace=pymupdf.csGRAY)
+            bw = pix.is_monochrome
+
+            results.append(bw)
+            label = "black & white" if bw==1 else "has color"
+            print(f"Page {i + 1}/{len(doc)}: {label}")
+
+        doc.close()
```
julian-smith-artifex-com: minor quibble - it's better to do:
```
with pymupdf.open(path) as doc:
    for i, page in enumerate(doc):
        ...
return results
```
which will automatically call `doc.close()`, even if an exception is raised.

julian-smith-artifex-com (approved):

docs/the-basics.rst
jamie-lemon: Updated

JorjMcKie (commented):
The wording ...
> A simple check to determine if a page is black and white or has color can be done by checking for whether a page is monochrome.

... is logically incorrect, because the negation of "black & white" is not the same as "page has colors".
A page with many "shades of gray/grey" is also not black & white but neither shows colors.

Black & white means that everything on the page is either pure white or pure black. **This** is what we are actually checking. If the result is `True` then we can conclude that the page has definitely no colors. But in case of `False` we cannot conclude that there are colors.

A page without colors is one where all pixels have R=G=B when the page was rendered as RGB.

Suggestion:

> A simple check to determine if a page is pure black and white ("monochrome"). Non‑monochrome means the page has multiple gray levels or colors.

PR #5040, opened 2026-07-06
Docs: updates converting files info & updates UTM links in header.

jamie-lemon:


docs/converting-files.rst
```diff
@@ -11,7 +11,7 @@ Converting Files
 Files to PDF
 ~~~~~~~~~~~~~~~~~~
 
-:ref:`Document types supported by PyMuPDF <HowToOpenAFile>` can easily be converted to |PDF| by using the :meth:`Document.convert_to_pdf` method. This method returns a buffer of data which can then be utilized by |PyMuPDF| to create a new |PDF|.
+:ref:`Document types supported by PyMuPDF <HowToOpenAFile>` can easily be converted to |PDF| by using the :meth:`Document.save` method after they have been opened.
```
julian-smith-artifex-com: Not sure whether it matters, but `convert_to_pdf()` returns a bytes, so is easier to use if one wants to open as a new pdf document. Where as save() always safes to a file.

docs/converting-files.rst
```diff
@@ -23,24 +23,18 @@ Files to PDF
...
 
 .. _Markdown_to_PDF:
 
 Markdown to PDF
 ~~~~~~~~~~~~~~~~~
 
-As Markdown files are supported input files they can be easily converted to PDF using the :meth:`Document.convert_to_pdf` method.
-
-In the simplest case you can just open the Markdown file and call the method to get a PDF representation of the content. 
+As Markdown files are supported input files they can be easily converted to PDF using the :meth:`Document.save` method. In the simplest case you can just open the Markdown file and call the method to get a PDF representation of the content.
```
julian-smith-artifex-com: One doesn't get a "PDF representation" of the content from save() in the way that `convert_to_pdf()` returnes.  - save() writes to a file and returns None, instead of returning a bytes.

docs/converting-files.rst
```diff
@@ -70,8 +64,7 @@ For example, to make all ``h1`` headers red (The single ``#`` symbol in Markdown
     css = "h1 {color:red;}"
     md_doc.apply_css(css)
 
-    pdf_doc = pymupdf.open(stream=md_doc.convert_to_pdf())
-    pdf_doc.ez_save("red-colored-header.pdf")
+    md_doc.save("red-colored-header.pdf")
```
julian-smith-artifex-com: You could use ez_save() here - it behaves in the same way as save(), converting to pdf if necessary.

JorjMcKie: As comfortable as the automatic PDF conversion now is - since we have integrated it in `save()` and friends - it is still too late (or early?) to make more adjustments to the **_PDF_** that's being created here.
Especially MD &#x21D2; PDF conversions profit a lot (often by an order of magnitude) from `subset_fonts()` before the save.
Therefore, we might want to consider adding `subset_fonts=True` as an additional `save`-option - and / or create a variant of `convert_to_pdf()` that directly returns a `Document` instead of `bytes`.
* `convert_to_pdfdoc()`?
* `convert_to_pdf(..., doc=True)`?

JorjMcKie (approved):

docs/converting-files.rst
jamie-lemon: Have added this kind of info back in

julian-smith-artifex-com (approved):

PR #5043, opened 2026-07-11
Prevent Death of None

JorjMcKie:
Prevent dropping refcount of Py_None to 0.


JorjMcKie:
@julian-smith-artifex-com - the test script will fail under Python versions < 3.12 without this fix, and will never (even without this fix) fail in versions 3.12+ because of the support of "immortal" objects - `Py_None` became immortal then.

julian-smith-artifex-com (approved):

PR #5072, opened 2026-07-30
Replace removed Python 2 C API macros for SWIG 4.5.0 compatibility

mjg:
SWIG was carrying thus macros before version 4.5.0, but they were never necessary in Python 3. Stop using for compatibility with upcoming SWIG 4.5.0.

Suggested-by: Jitka Plesnikova <jplesnik@redhat.com>

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -21945,7 +21945,7 @@ def JM_set_widget_properties(annot, Widget):
 
     # entry ignored - may be used later
     #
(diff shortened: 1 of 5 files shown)


mjg:
Apparently, one hunk has been applied in `d2ddf903 ("src/extra.i: fix for swig-4.5.", 2026-08-22)`. Is this enough, though?

PR #5074, opened 2026-08-02
Fix leaked reference in _as_blocks()

### LINKED ISSUE #5067: Memory leak in `extractBLOCKS` (`_as_blocks`): text object reference never released
## Description

The `extractBLOCKS` code path leaks a Python object for **every** text, image, and vector block on **every** call. Since `extractBLOCKS` (a.k.a. `page.get_text("blocks")`) is commonly called once per page in bulk-processing loops, the leak accumulates quickly across large/multi-page documents.

## Location

`src/extra.i`, in the recursive helper `_as_blocks` (around line 3501).

## Root cause

The block tuple is populated like this:

```c
PyObject *litem = PyTuple_New(7);
...
PyTuple_SET_ITEM(litem, 4, Py_BuildValue("O", text));   // <-- leak
...
```

`text` is created as a **new reference**, via either:

- `JM_EscapeStrFromBuffer(res)` (text blocks), or
- `PyUnicode_FromFormat(...)` (image / vector blocks).

The `"O"` format code in `Py_BuildValue` **increments** the refcount of `text` and returns a *separate* new object. That separate object is the one stolen into the tuple by `PyTuple_SET_ITEM`. The original `text` reference is then discarded on the next iteration:

```c
text = NULL;
```

…without ever being decref'd. Net effect: `text` is leaked once per block.

The other six `PyTuple_SET_ITEM` calls in the same block are correct — `Py_BuildValue("f"/"i", ...)` creates a fresh object the tuple legitimately owns.

## Fix

Hand the *original* `text` reference to the tuple instead of creating an extra one. Since `PyTuple_SET_ITEM` steals a reference, set it directly:

```c
PyTuple_SET_ITEM(litem, 4, text);
```

(Equivalently, `Py_BuildValue("N", text)`, where `"N"` steals rather than increments.)

## Environment

- PyMuPDF: `<fill in version, e.g. output of fitz.__doc__>`
- OS / Python: `<fill in>`

## Reproduction

```python
import pymupdf, os, psutil
doc = pymupdf.open("some_multipage.pdf")
p = psutil.Process(os.getpid())
for i in range(200):
    for page in doc:
        page.get_text("blocks")
    rss_mb = p.memory_info().rss / (1024 * 1024)  # Convert to MiB
    print(f"Iteration {i + 1}: {rss_mb:.2f} MiB")  # grows steadily
```

### PyMuPDF version

1.27.2.2

### Operating system

Linux

### Python version

3.12

KetanHegde:
## Summary

Fixes #5067.

`JM_EscapeStrFromBuffer()` and `PyUnicode_FromFormat()` return new references. In `_as_blocks()`, `text` was wrapped with `Py_BuildValue("O", text)` before being passed to `PyTuple_SET_ITEM()`. The `"O"` format increments the reference count, while `PyTuple_SET_ITEM()` steals the reference it receives, leaving the original reference to `text` unreleased.

This change passes `text` directly to `PyTuple_SET_ITEM()`, transferring ownership to the tuple and avoiding the leaked reference.

## Testing

- Built PyMuPDF successfully after the change.
- Ran the existing test suite successfully.

### DIFF
--- a/src/extra.i
+++ b/src/extra.i
@@ -3498,7 +3498,7 @@ int _as_blocks(fz_stext_block *block, fz_rect tp_rect, PyObject *lines, int bloc
             PyTuple_SET_ITEM(litem, 1, Py_BuildValue("f", blockrect.y0));
             PyTuple_SET_ITEM(litem, 2, Py_BuildValue("f", blockrect.x1));
             PyTuple_SET_ITEM(litem, 3, Py_BuildValue("f", blockrect.y1));
-            PyTuple_SET_ITEM(litem, 4, Py_BuildValue("O", text));
+            PyTuple_SET_ITEM(litem, 4, text);
             PyTuple_SET_ITEM(litem, 5, Py_BuildValue("i", block_n));
             PyTuple_SET_ITEM(litem, 6, Py_BuildValue("i", block->type));
             LIST_APPEND(lines, litem);


KetanHegde:

KetanHegde:
recheck

julian-smith-artifex-com:
I've rebased this PR on top of latest main, which should fix the test failure with mupdf master.

julian-smith-artifex-com (approved):

julian-smith-artifex-com:
Thanks for this, looks good, so merging now.

PR #5076, opened 2026-08-03
Fix several small correctness bugs

A831ARD0:
## Summary

A batch of small, independent correctness fixes found while reviewing `src/__init__.py`, `src/utils.py`, `src/_apply_pages.py`, and the table-extraction module. Each item below is a separate, self-contained fix; happy to split into individual PRs if preferred.

- **`Archive.add()`**: a missing comma in an `assert` turned the error message into part of the boolean expression, silently disabling the empty-name check for binary content (`Archive().add(b"...", path="")` no longer raised).
- **`Annot.file_info`**: `UnboundLocalError` when a malformed/crafted attachment's `/FS` dict has neither `/UF` nor `/F`.
- **`Page.set_language()`**: passed the bound method `pdfpage.obj` instead of calling it (`pdfpage.obj()`), breaking every call that actually sets a language.
- **`Widget.on_state()`**: `TypeError` when `button_states()` leaves an entry as `None`.
- **`Pixmap.tobytes()`**: missing f-string prefix meant the "cannot have alpha" error literally printed `'{output}'` instead of the format name.
- **`JM_matrix_from_py()`**: copy-paste bug returning `mupdf.FzRect()` instead of `mupdf.FzMatrix()` on the float-conversion failure path (latent — every other sibling converter returns its own correct type).
- **`utils.get_text()`**: a bare `assert option in formats` ran before the following `if`, making the documented "misspelled option falls back to `'text'`" behavior unreachable (and behavior differed under `-O`).
- **`utils.get_label_pno()`**: raised `IndexError` instead of returning `""` (as documented) when no label rule applies to the given page number.
- **`utils.get_sorted_text()`**: `ZeroDivisionError` for a zero-width glyph bbox.
- **`_apply_pages.py` (`method='fork'`)**: workers received plain list indices (`range(len(pages))`) instead of the actual page numbers, so `apply_pages(path, fn, pages=[3, 7, 9], method='fork')` silently processed pages 0, 1, 2 instead — results were silently attributed to the wrong pages. Fixed by passing `(index, page_number)` pairs through the queue.
- **`table.chars_in_rect()`**: compared a top-down rect (from `page.get_drawings()`) against `c["y0"]`/`c["y1"]`, which are in PDF-native (bottom-up) space, instead of the top-down `c["top"]`/`c["bottom"]`. This made box-drawn table borders (filled/connected rectangles instead of ruling lines) fail to register as table candidates.

## Testing

I don't have a build of the compiled `_fitz`/mupdf C extension available, so I could not run the full test suite. Each fix was verified by:
- Reading the surrounding code and confirming the defect against the function's own documented contract or a directly comparable sibling function.
- `python3 -m py_compile` on every modified file to confirm no syntax errors were introduced.

Given that, please review carefully and let me know if you'd like any of these split out, reverted, or backed by a regression test before merging.

### DIFF
--- a/src/_apply_pages.py
+++ b/src/_apply_pages.py
@@ -137,11 +137,12 @@ def _fork(
         while 1:
             if verbose:
                 pymupdf.log(f'{os.getpid()=}: calling get().')
-            page_num = queue_down.get()
+            item = queue_down.get()
             if verbose:
-                pymupdf.log(f'{os.getpid()=}: {page_num=}.')
-            if page_num is None:
+                pymupdf.log(f'{os.getpid()=}: {item=}.')
+            if item is None:
                 break
+            index, page_num = item
             try:
                 if not document:
                     if stats:
@@ -172,9 +173,9 @@ def _fork(
                 if verbose: pymupdf.log(f'{os.getpid()=}: exception {e=}')
                 ret = e
             if verbose:
-                pymupdf.log(f'{os.getpid()=}: sending {page_num=} {ret=}')
-                
-            queue_up.put( (page_num, ret) )
+                pymupdf.log(f'{os.getpid()=}: sending {index=} {ret=}')
+
+            queue_up.put( (index, ret) )
 
     error = None
 
@@ -206,8 +207,8 @@ def _fork(
             t = time.time()
         if verbose:
             pymupdf.log(f'Sending page numbers.')
-        for page_num in range(len(pages)):
-            queue_down.put(page_num)
+        for index, page_num in enumerate(pages):
+            queue_down.put((index, page_num))
         if stats:
             _stats_write(t, 'Send page numbers')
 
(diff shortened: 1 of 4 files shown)


A831ARD0:

src/_apply_pages.py
```diff
@@ -137,11 +137,12 @@ def childfn():
...
                 pymupdf.log(f'{os.getpid()=}: calling get().')
-            page_num = queue_down.get()
+            item = queue_down.get()
             if verbose:
-                pymupdf.log(f'{os.getpid()=}: {page_num=}.')
-            if page_num is None:
+                pymupdf.log(f'{os.getpid()=}: {item=}.')
+            if item is None:
                 break
+            index, page_num = item
```
julian-smith-artifex-com, line 145: I think the current code is ok here. This code is in the worker function `def childfn():` so reads page numbers from queue_down and writes (page_num, text) to queue up.

src/_apply_pages.py
```diff
@@ -172,9 +173,9 @@ def childfn():
                 if verbose: pymupdf.log(f'{os.getpid()=}: exception {e=}')
                 ret = e
             if verbose:
-                pymupdf.log(f'{os.getpid()=}: sending {page_num=} {ret=}')
-                
-            queue_up.put( (page_num, ret) )
+                pymupdf.log(f'{os.getpid()=}: sending {index=} {ret=}')
+
+            queue_up.put( (index, ret) )
```
julian-smith-artifex-com, line 178: As above, this is worker code.

src/utils.py
```diff
@@ -497,7 +497,6 @@ def get_text(
         "blocks": pymupdf.TEXTFLAGS_BLOCKS,
     }
     option = option.lower()
-    assert option in formats
     if option not in formats:
         option = "text"
```
julian-smith-artifex-com: I think the intention here is to fail if an unrecognised option is specified, so we should remove the `if option not in formats: option = "text"` code, rather than the assert.

julian-smith-artifex-com:
Thanks for submitting this PR, looks like some good fixes in the areas i know about. I've enabled the pre-push tests.

You should be able to build/test fairly easily if you run `pip` in a python venv - all required packages etc will be installed automatically.

PR #5088, opened 2026-08-16
Chore: consistent metadata in wheels

dimbleby:
I do not fully understand why `pipcl` and `pytest` become project dependencies on pyodide only, but the best way to express that is through environment markers.  Else, wheels uploaded to pypi contain different metadata.  

Cross-platform resolvers like uv and poetry typically just look at one wheel and trust that all have the same requirements: so depending on which they happen to look at: you are either getting pipcl and pytest always installed, or never installed.

Signaling the requirement through static metadata allows the right answer to be reached on all platforms.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1105,11 +1105,12 @@ version_mupdf = '1.28.2'
 
 # A normal PyMuPDF package.
 
-requires_dist = list()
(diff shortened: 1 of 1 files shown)


dimbleby:
I will not be signing a CLA, you are welcome to this code, take it or leave it

PR #5095, opened 2026-08-20
Fix ignoring aspect ratio for square images

JorjMcKie:
Fix an corner case:
Keeping aspect ratio is ignored when inserting a square image into a non-square target rectangle.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -22669,7 +22669,11 @@ def calc_image_matrix(width, height, tr, rotate, keep):
         f = fw
         fw = fh
         fh = f
-    if fw < 1:
+    if keep:
+        side = min(trw, trh)
+        w = side
+        h = side
+    elif fw < 1:
         if trw / fw > trh / fh:
             w = trh * small
             h = trh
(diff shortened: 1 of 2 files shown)


src/__init__.py
```diff
@@ -22683,6 +22683,10 @@ def calc_image_matrix(width, height, tr, rotate, keep):
         else:
             w = trw
             h = trw * small
+    elif keep:
+        side = min(trw, trh)
+        w = side
+        h = side
```
julian-smith-artifex-com: Could we put the new code as the first item in the if..else.. chain? Otherwise it's not clear that this branch will always be taken if `keep` is true.
I.e.:
```
if keep:
    side = min(trw, trh)
    w = side
    h = side
elif fw < 1: ...
elif fw != fh: ...
else: ...
```

JorjMcKie: sure

JorjMcKie: Done

julian-smith-artifex-com (approved):

PR #5096, opened 2026-08-20
Fix chars_in_rect pairing top-down rects with top/bottom

### LINKED ISSUE #5092: [BUG] Coordinate key mismatch in `pymupdf.table.chars_in_rect()` drops table candidates and rows
### Description of the bug

In `pymupdf/table.py`, the internal helper function `chars_in_rect(CHARS, rect)` compares a top-down bounding box `rect` against character coordinates in `CHARS`.

However, while `rect` uses top-down Y coordinates (`rect[1]` is top, `rect[3]` is bottom) and the horizontal check uses top-down `x0`/`x1`, the vertical check in `chars_in_rect` compares `rect[1]` and `rect[3]` against `c["y0"]` and `c["y1"]` instead of `c["top"]` and `c["bottom"]`.

```python
def chars_in_rect(CHARS, rect):
    return any(
        1
        and rect[0] <= c["x0"]
        and c["x1"] <= rect[2]
        and rect[1] <= c["y0"]    # ❌ rect[1] is top-down Y, but c["y0"] is CTM / bottom-up
        and rect[3] >= c["y1"]    # ❌ rect[3] is top-down Y, but c["y1"] is CTM / bottom-up
        for c in CHARS
    )
```

**Impact on Table Detection (`find_tables()`):**
In `table.py::clean_graphics()`, candidate table boundary rectangles (`prect0`) are filtered using `chars_in_rect(CHARS, prect0)` to determine if a candidate box contains text. 
* On standard full-page tables, `c["y0"]` and `c["top"]` happen to numerically overlap by coincidence, masking the issue.
* On short tables near the bottom of a page, continuation tables across page breaks, or pages with non-identity CTMs, `c["y0"]`/`c["y1"]` do not overlap with top-down `rect[1]`/`rect[3]`. 
* As a result, `chars_in_rect` returns `False`, `clean_graphics()` discards the candidate boundary box, line synthesis fails, and entire table rows (or short tables) disappear from `find_tables()`.

Note that elsewhere in `table.py` (e.g. in `has_text()` around line 1554), top-down coordinates are correctly paired with `c["top"]` and `c["bottom"]`:
```python
text_bboxes = [
    (c["x0"], c["top"], c["x1"], c["bottom"]) # ✅ Correct top-down pair used in has_text()
    for c in CHARS
    if c["text"] not in white_spaces
]
```

#### Minimal Code Example 1 (Direct Unit Test):
```python
import pymupdf.table

# A character positioned at top=100.0, bottom=110.0 on an 800px page.
# Top-down coordinates: top=100, bottom=110
# CTM / bottom-up coordinates: y0=690, y1=700
char = {
    "x0": 10.0, "x1": 20.0,
    "top": 100.0, "bottom": 110.0,
    "y0": 690.0, "y1": 700.0
}

# Top-down bounding box clearly surrounding the character (top=90, bottom=120)
rect = (0.0, 90.0, 30.0, 120.0)

# Expected: True (character is inside the top-down bounding box)
# Actual: False (because chars_in_rect compares top-down rect bounds with y0/y1)
result = pymupdf.table.chars_in_rect([char], rect)
print("chars_in_rect result:", result)  # Outputs: False
```

#### Proposed Fix:
In `src/table.py`, update `chars_in_rect` to use `c["top"]` and `c["bottom"]`:

```diff
def chars_in_rect(CHARS, rect):
    """Check whether any of the chars in CHAR are inside rectangle 'rect'."""
    return any(
        1
        and rect[0] <= c["x0"]
        and c["x1"] <= rect[2]
-       and rect[1] <= c["y0"]
-       and rect[3] >= c["y1"]
+       and rect[1] <= c["top"]
+       and rect[3] >= c["bottom"]
        for c in CHARS
    )
```

### PyMuPDF version

1.27.2.3

### Operating system

MacOS

### Python version

3.12

BetterAndBetterII:
## Summary

- `chars_in_rect()` compared a top-down `rect` (`rect[1]` top, `rect[3]` bottom) against CTM `y0`/`y1`. Short tables and non-identity CTMs then failed the text-in-rect check and were dropped from `find_tables()`.
- Pair `rect[1]`/`rect[3]` with `c["top"]`/`c["bottom"]`, matching `has_text()` in the same file.
- Fixes #5092

## Test plan

- [x] RED then GREEN: synthetic char with `top`/`bottom` inside `rect` and `y0`/`y1` outside is now detected
- [x] `tests/test_tables.py` — `test_5092` and `test_table2` passed

### DIFF
--- a/src/table.py
+++ b/src/table.py
@@ -200,8 +200,8 @@ def chars_in_rect(CHARS, rect):
         1
         and rect[0] <= c["x0"]
         and c["x1"] <= rect[2]
-        and rect[1] <= c["y0"]
-        and rect[3] >= c["y1"]
+        and rect[1] <= c["top"]
+        and rect[3] >= c["bottom"]
         for c in CHARS
     )
 
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -939,3 +939,23 @@ def test_find_tables_union_no_layout_degrades_to_line_candidates():
     finally:
         pymupdf._get_layout = original_get_layout_fn
         doc.close()
+
+
+def test_5092():
+    """chars_in_rect must pair a top-down rect with top/bottom, not CTM y0/y1.
+
+    A character whose top/bottom sit inside rect, but whose y0/y1 do not,
+    must still be detected. This matches the has_text() pairing.
+    """
+    from pymupdf.table import chars_in_rect
+
+    char = {
+        "x0": 10.0,
+        "x1": 20.0,
+        "top": 100.0,
+        "bottom": 110.0,
+        "y0": 690.0,
+        "y1": 700.0,
+    }
+    rect = (0.0, 90.0, 30.0, 120.0)  # top-down: top=90, bottom=120
+    assert chars_in_rect([char], rect)


BetterAndBetterII:

veget-able:
Thanks for the fix! 

I also tested this against our 503-page table benchmark.
The change looks correct to me. The overall GTRM moved only slightly, from 0.721066 to 0.720618, with just 4 of 503 pages changing.

Looking at those pages individually, one improved significantly, one regression appears to be a ground-truth mismatch, and the remaining two exposed pre-existing edge cases in the refine pipeline that were previously masked because these candidates were being rejected.

So I don't think the small benchmark regression is a reason to hold this fix.

veget-able:
@JorjMcKie 
It looks like the CLA check is incomplete because @cursoragent is included as a committer and cannot sign it.

We may run into this more often with agent-assisted PRs, so I think it would be worth deciding how we want to handle agent accounts in the CLA process going forward.

veget-able (approved):

julian-smith-artifex-com:
There's a cla-assistant issue for this: https://github.com/cla-assistant/cla-assistant/issues/1190, but it only has two comments (including my new one).

PR #5097, opened 2026-08-20
Add support for field hierarchies

JorjMcKie:
This fix adds Form Field hierarchy support to our Widget handling.
Background:

PyMuPDF (like MuPDF itself) always inserts new Form Fields as widgets on the respective page. All inserted widgets automatically become root fields in the PDF.
There is no way to add complex information structures like `Person.Name.First` such that the widget only has the name part `First` and the other segments are parents in multiple hierarchy levels.

This fix solves this by automatically adding the hierarchy where needed. For the example above, a chain `Person → Name → First` will be created. `Person` becomes a root field pointing to `Name` via its `/Kids` array, and similarly `Name` points to `First`. `Name` and `First` in turn will point backwards to their parents via the `/Parent` key. This allows for tree-like information structures like this:
```
Person
  ↪ Name
     ↪ First
     ↪ Last
  ↪ Address
     ↪ Street
     ↪ City
```
The only root field would be `Person`. The four level 3 items would represent widgets living on any pages.
This behavior is triggered by the presence of the name segmentation indicator colon ".".
If no colon is present, the old behavior occurs.

### DIFF
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -4,6 +4,7 @@ Test PDF field (widget) insertion.
 """
 import pymupdf
 import os
+from pymupdf import mupdf
 
 scriptdir = os.path.abspath(os.path.dirname(__file__))
 filename = os.path.join(scriptdir, "resources", "widgettest.pdf")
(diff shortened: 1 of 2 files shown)


julian-smith-artifex-com:
[mupdf master and 1.28.x have just been updated with the fix for swig-4.5. We need to rerun test_quick after these changes have made it to the github mupdf mirror, so rerun in an hour or so.]

tests/test_widgets.py
```diff
@@ -486,3 +487,51 @@ def test_4950():
...
+        """Check the names climbing up the hierarchy."""
+        xref = w.xref
+        parts = [""] + w.field_name.split(".")
+        obj = mupdf.pdf_load_object(pdfdoc, xref)  # the widget object
+        for part in reversed(parts):
+            obj = obj.pdf_dict_get(pymupdf.PDF_NAME("Parent"))
+            if obj is None or obj.pdf_is_null():
+                break
+            pdf_name = obj.pdf_dict_get(pymupdf.PDF_NAME("T")).pdf_to_string()
+            assert pdf_name[0] == part
```
julian-smith-artifex-com: This seems to allow early termination of the loop with no error?

Could we assert that each item in `parts` is matched in the loop, and then assert that the next return from `obj = obj.pdf_dict_get(pymupdf.PDF_NAME("Parent"))` is None?

JorjMcKie: The condition `obj is None or obj.pdf_is_null()` is not an error but indicates that we have reached the top of the tree, i.e. `AcroForm/Fields`. Objects in the Fields array by definition have no `/Parent`. We therefore must skip the following check.
I should have commented this properly.

JorjMcKie: Or simply this:
```python
if not obj.pdf_is_null():
    pdf_name = obj.pdf_dict_get(pymupdf.PDF_NAME("T")).pdf_to_string()
    assert pdf_name[0] == part
```

JorjMcKie: It seems you have this type of check in mind:
```python
    def chain_upwards(w):
        """Check the names climbing up the hierarchy."""
        xref = w.xref
        parts = [""] + w.field_name.split(".")
        obj = mupdf.pdf_load_object(pdfdoc, xref)  # the widget object
        part_count = len(parts)
        for i, part in enumerate(reversed(parts)):
            obj = obj.pdf_dict_get(pymupdf.PDF_NAME("Parent"))
            if not (obj is None or obj.pdf_is_null()):
                pdf_name = obj.pdf_dict_get(pymupdf.PDF_NAME("T")).pdf_to_string()
                assert pdf_name[0] == part
            else:  # assert we have reached the top of the hierarchy
                assert i == part_count - 1
```

julian-smith-artifex-com (approved):

PR #5102, opened 2026-08-25
Document Widget Support Updates

JorjMcKie:
Updates the documentation of new features around PDF Form Fields:

* Support of form field hierarchies
* Support for widget aliases
* Support for RBGs - radio button groups. RBGs are now automatically created or extended when a widget with the same name is added to a page.


jamie-lemon (approved):
LGTM!

PR #5127, opened 2026-09-21
Improvements to Widget handling

JorjMcKie:
* Improved / streamlined Widget insertion
* More complete  Widget deletion: We now automatically also remove parent fields when their /Kids arrays become empty.

add test script


src/__init__.py
```diff
@@ -10205,40 +10209,50 @@ def _adjust_widget(self, widget):
...
+            PDF_NAME("MaxLen"),
+            PDF_NAME("Q"),
+            PDF_NAME("DS"),
+            PDF_NAME("DR"),
+            PDF_NAME("TI"),
+            PDF_NAME("RV"),
+        )
+        for key_name in MOVE_KEYS:
+            new = annot_obj.pdf_dict_get(key_name)
+            old = parent_obj.pdf_dict_get(key_name)
```
julian-smith-artifex-com: Looks like "old" and "new" are reversed from previous code? Could we rename these to be "widget" and "parent" for clarity? (Or maybe there are better names.)

src/__init__.py
```diff
@@ -11307,14 +11321,99 @@ def finished():
...
-        page.delete_annot(annot)
+
+        # --- 1) Remove widget from page /Annots array ---
+        annots = page_obj.pdf_dict_get(PDF_NAME("Annots"))
+        annots_count = annots.pdf_array_len()
+
+        for i in reversed(range(annots_count)):
+            a = annots.pdf_array_get(i)
+            if a.pdf_to_num() == wxref:
+                annots.pdf_array_delete(i)
```
julian-smith-artifex-com: Looks like there are 5 occurrencies of this loop; could we move it into a separate local function?

JorjMcKie: Good point ...

src/__init__.py
JorjMcKie: Ok, changed "new" to "widget_itm" and "old" to "parent_itm".

tests/test_widgets.py
```diff
@@ -312,14 +312,15 @@ def test_interfield_calculation():
...
-        # the xref of the i-th item must equal that of the last widget
-        assert (
-            pymupdf.mupdf.pdf_to_num(pymupdf.mupdf.pdf_array_get(CO, i))
-            == list(page.widgets())[-1].xref
-        )
+        # last item in the /CO array is that of the last widget's parent
+        last_xref = CO.pdf_array_get(i).pdf_to_num()
+        obj = pymupdf.mupdf.pdf_load_object(pdf, w.xref)
+        parent_xref = obj.pdf_dict_get(pymupdf.PDF_NAME("Parent")).pdf_to_num()
+        assert last_xref == parent_xref
```
julian-smith-artifex-com: Is this code identical in behaviour to the previous code?

JorjMcKie: yes: it does the same checks from a logical point of view

julian-smith-artifex-com (approved):

